PR #34, opened 2020-06-07
v2.0.0

exelban:
- rewritten application from scratch
- new Settings
- new custom popup view
- moved to own implementation of chart
- added more option to configure a widget
- now each module has own widget in menu bar
- a lot of new features...


ModuleKit/module.swift
```diff
@@ -0,0 +1,287 @@
...
+        get {
+            let widgetStr = self.store?.pointee.string(key: "\(self.config.name)_widget", defaultValue: self.config.defaultWidget.rawValue)
+            return widget_t.allCases.first{ $0.rawValue == widgetStr } ?? widget_t.unknown
+        }
+        set {}
+    }
+    private var ready: Bool = false
+    private var widgetLoaded: Bool = false
+    
+    public init(store: UnsafePointer<Store>?, popup: NSView?, settings: Settings_v?) {
```
dstd, line 86: Could you explain me, please, what was the reason to inject this dependency as a UnsafePointer? `weak` modifier seems to be enough here

dstd: Hmm.. Actually you're passing the pointer to the variable that contains a reference to `Store` object, and every `Module` contains sucha indirect reference.

This could be useful if `Store` object is supposed to be replaced with new one on the fly, so that after the change of the value in `store: Store` variable (in AppDelegate.swift) all `Module`s got new `Store` with no additional effort.

IMHO, all these `store: UnsafePointer<Store>` could be replaced with `weak var store: Store?`.

Although I proposed `weak` to keep original retain-free behavior, looks like it is not necessary too, because `store` variable in `AppDelegate` is actually immutable so that a lifetime of `Store` equals to a lifetime of the application.


To bottom line it:
```diff
AppDelegate.swift
- var store: Store = Store()
+ let store: Store = Store()
```
```diff
module.swift 
- private var store: UnsafePointer<Store>
+ private let store: Store

- self.enabled = self.store.pointee.bool(key: ...
+ self.enabled = self.store.bool(key: ...
```
```diff
readers.swift
- public var store: UnsafePointer<Store>? = nil
+ public var store: Store?

- let removableState = store?.pointee.bool(key: ...
+ let removableState = store?.bool(key: ...
```
and so on

exelban: Thanks for the explanation. Yes, I suppose pointers could be removed at all.

PR #42, opened 2020-06-24
add Homebrew install instruction

dawidd6:
Added to Homebrew Cask repo here: https://github.com/Homebrew/homebrew-cask/pull/84790

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ Simple macOS system monitor in your menu bar
 ## Installation
 You can download latest version [here](https://github.com/exelban/stats/releases).
 
+### Homebrew
+
+```bash
+brew cask install stats
+```
+
 ## Requirements
 
 Stats is currently supported on macOS 10.14 (Mojave) and higher.


exelban:
@dawidd6 thanks for this. 

As I understand homebrew fetch a new version from git tags?
So I don't need to update anything to provide a new version for brew?

dawidd6:
Your program has an auto-update mechanism built-in, so no need to worry about it.

exelban:
Yeap, but I see that they have a `version` inside. And fetching the app from `https://github.com/exelban/stats/releases/download/v#{version}/Stats.dmg`. 

So it means, if I release a new version tomorrow, I will need to update their configuration file. To ensure that users install the latest app via homebrew?

Or it will fetch a new version if there is a new tag in git?

exelban:
Because if they just fetch a specific tag. And not checking for the newest tag. It will be good to add a `latest` tag to git. And use this tag to indicate the last version.

dawidd6:
Yes, the Cask file should be updated, when a new git tag is pushed, but I think it is not too important, because auto-update in the app will always get it to the newest version regardless.

exelban:
Auto-update in Stats app?

dawidd6:
Yes

exelban:
Hmmm, it's not a good look, when someone installs the new app. Open it, and receive the message that a new version is available.

I see two options: create a gh action to make automatically PR to brew, or add `latest` tag.

But still thanks for your help.

dawidd6:
I understand your concern. There is an Action for Homebrew Formulae updating, but sadly not for Casks (or I'm not aware of any).

PR #110, opened 2020-10-12
New app icon

jaykovski:


exelban:
Thanks for this. It looks really nice.
Could you send me an original project of this icon? It might be needed in the future.
Send it to my email please `mitrovtsiy@ukr.net`.

PR #127, opened 2020-10-26
Adding Spanish translation and translator info

santius:
Added Spanish as language in Readme.md file


exelban:
Thanks, but I will do it myself. I add a new language to the end of the list (regarding the date when lang was added).

Issue #118, opened 2020-10-16
[BUG] Wrong network speed


xfzka:
**Describe the bug**
Wrong network upload speed 

<img width="552" alt="image" src="https://user-images.githubusercontent.com/24959175/96290315-faec9400-1018-11eb-8aa2-e110edd1aa18.png">

<img width="832" alt="image" src="https://user-images.githubusercontent.com/24959175/96290387-122b8180-1019-11eb-8695-a75114bec8f2.png">

This app is show total network traffic to upload speed.
When i upload a 3MB size file, the upload speed +3MB, upload 100MB size file, the upload speed +100MB.
And now the upload speed is 3.8GB/s

<img width="212" alt="image" src="https://user-images.githubusercontent.com/24959175/96291542-c7126e00-101a-11eb-8c28-c9457ac66371.png">

And I also found that when the speed reaches 4GB/s or large than 4GB/s the speed will down to the zero, and then re-accumulate.

**Details:**
 - MacOS 10.15.7 
 - Application version 2.3.7 (2.3.6 also has this bug, I am use this app start  from 2.3.6 version)

exelban:
Hi.

Did you use the internal network adapter or some thunderbolt adapter?

xfzka:
I use internal network adapter and wifi ,  internal network adapter first.

exelban:
So it's an internal wifi adapter?

It's strange behavior. I just read the value from the system. There is no place where I can do the wrong calculation or something.

HartwigHarder:
I do have the same issue, using a USB dock, but the network speed in the system overview is reasonable. 
Stats does show the correct adapter and network usage from the top process list when opening the stats menu.

exelban:
Hi.

I know that there is a problem with external interfaces. 
It looks like I need to find some new way to read the transferred data size.

exelban:
Fixed in the [last](https://github.com/exelban/stats/releases/tag/v2.4.5) release.

PR #157, opened 2020-11-13
Fix: No download speed

AhmedOS:
### Summery
- This PR should fix issue https://github.com/exelban/stats/issues/118.
- This will be achieved by summing download speeds from all processes.
- This approach will be triggered automatically as an alternative approach if regular one didn't work after three reading attempts.

### DIFF
--- a/Modules/Net/readers.swift
+++ b/Modules/Net/readers.swift
@@ -99,7 +99,7 @@ internal class UsageReader: Reader<Network_Usage> {
         freeifaddrs(interfaceAddresses)
         
         if !verifiedUsageDataApproach {
-            if nonZeroUpdatesCount >= 3 {
+            if didNativeApproachReportNonZeroDownload || nonZeroUpdatesCount >= 3 {
                 if !didNativeApproachReportNonZeroDownload {
                     shouldUseProcessesUsageData = true
                 }
--- a/Modules/Net/readers.swift
+++ b/Modules/Net/readers.swift
@@ -136,69 +136,8 @@ internal class UsageReader: Reader<Network_Usage> {
         self.usage.download = accumulatedDownload
     }
     
-    private func allProcessesTotalUsage() -> Network_Usage? {
-        let task = Process()
-        task.launchPath = "/usr/bin/nettop"
-        task.arguments = ["-P", "-L", "1", "-k", "time,interface,state,rx_dupe,rx_ooo,re-tx,rtt_avg,rcvsize,tx_win,tc_class,tc_mgt,cc_algo,P,C,R,W,arch"]
-        
-        let outputPipe = Pipe()
-        let errorPipe = Pipe()
-        
-        task.standardOutput = outputPipe
-        task.standardError = errorPipe
-        
-        do {
-            try task.run()
-        } catch let error {
-            print(error)
-            return nil
-        }
-        
-        let outputData = outputPipe.fileHandleForReading.readDataToEndOfFile()
-        let errorData = errorPipe.fileHandleForReading.readDataToEndOfFile()
-        let output = String(decoding: outputData, as: UTF8.self)
-        _ = String(decoding: errorData, as: UTF8.self)
-        
-        if output.isEmpty {
-            return nil
-        }
-
-        var list: [Network_Process] = []
-        var firstLine = false
-        output.enumerateLines { (line, _) -> () in
-            if !firstLine {
-                firstLine = true
-                return
-            }
-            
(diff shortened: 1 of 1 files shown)


exelban:
Hi. 

Do you have this type of bug (#118)? Because I don't. And cannot check if it's fixing it.
Also, I suppose #156 can be closed if this will be merged?

AhmedOS:
Yes I have same issue, however, I noticed that processes download speed working fine, so the fix is based on that.
Yes right, https://github.com/exelban/stats/pull/156 should be closed if this PR is merged.

exelban:
Ok, thanks. I will take a look later today.

Modules/Net/readers.swift
```diff
@@ -93,22 +98,100 @@ internal class UsageReader: Reader<Network_Usage> {
...
+            }
+            
+            list.append(process)
+        }
+        
+        var usage = Network_Usage()
+        for process in list {
+            usage.upload += Int64(process.upload)
+            usage.download += Int64(process.download)
+        }
```
exelban: You don't need to have a second loop here. These calculations can be done in the `output.enumerateLines` loop.

Modules/Net/readers.swift
```diff
@@ -93,22 +98,100 @@ internal class UsageReader: Reader<Network_Usage> {
...
+        self.usage.download = max(accumulatedDownload - self.usage.download, 0)
+        
         self.callback(self.usage)
         
         self.usage.upload = upload
-        self.usage.download = download
+        self.usage.download = accumulatedDownload
+    }
+    
+    private func allProcessesTotalUsage() -> Network_Usage? {
```
exelban: Also, there is no need to have a `Network_Usage` struct here. I propose to return a tuple.

Modules/Net/readers.swift
```diff
@@ -93,22 +98,100 @@ internal class UsageReader: Reader<Network_Usage> {
...
-        }
-        if self.usage.download < 0 {
-            self.usage.download = 0
+        let accumulatedDownload: Int64
+        if shouldUseProcessesUsageData {
+            let processesUsage = allProcessesTotalUsage()
+            accumulatedDownload = processesUsage?.download ?? 0
+        } else {
+            accumulatedDownload = download
         }
```
exelban, line 125: I did not really understand why there are so many checks. But it still not checking properly.

Correct me please if I miss something.
`nonZeroUpdatesCount` checks if the last 3 reads were 0. But it normal that in the last second traffic was 0. And even more than the last 3 seconds. 

Also, this approach does not work if I switch to some interface that does not have traffic. And now it starts to showing some download values. And it can be like this.

exelban (commented):
Check please these notes.

exelban:
I think about how it can be implemented. And found only 1 good solution.

It's almost impossible to detect if the default reader does not work properly. Only the user can determine that something is not good. So. Maybe add the option in the settings to switch to the alternative reader (sum of the processes)?

Also, this way of reading the bandwidth will not work with the interface switcher. So I think will be good to add the option in the interface options to read the value based on processes not directly from the interface. Or just add a trigger `Alternative reader`. And interface selector must be unavailable (hidden) when this option is enabled.

AhmedOS:
Yes that approach won't consider selected interface, which is a drawback. However, having an option to trigger the alternative approach sounds better, so we won't rely on that ad-hoc automatic triggering, and it will be clear that reported bandwidth isn't from selected interface.
But I would suggest that activating the alternative approach for getting bandwidth won't disable selecting the preferred interface, to be able to get its details (IPs for example), instead, adding a separate option to select bandwidth reporting approach may be better, that is, reporting from selecting interface only or all processes usage.

There is another approach I thought about but I'm not sure if it is feasible: We may use the interface reported with processes list from `nettop` to filter for the selected interface, this will require removing `interface` from excluded columns argument, and removing `-P` also to get details not only processes summery. But the problem is that I'm not sure how to distinguish between detailed row and process summery one (so we can skip the summery from calculations), both of them have 5 items. Anyways, I think the former approach is better and less error-prone.

AhmedOS:
I have changed implementation so alternative approach will be triggered manually via `shouldReportSelectedInterfaceBandwidthOnly` flag. Remains implementing UI, and resetting usage when bandwidth approach is changed, feel free to continue from current progress.

exelban:
Thanks. I will take a look later this week.

exelban:
Thanks for your contribution. Sorry for the long acceptance.

Issue #170, opened 2020-11-18
Deleting wrong element while enumerating sensors


Military-Doctor:
**Describe the bug**
Here are a bug when you are trying to delete an sensor which temperature is higher than 110 Celsius.

Due to the deletion during the enumerating.

**Details:**
 - Device: [MacBook Pro 16'' 2019, i7-9750H, 16G]
 - MacOS [10.15.7 (19H15)]
 - Application version [2.4.3]

**Explain**
`Modules/Sensors/readers.swift`

It's a for loop around line 44, I added a line to print the list count to see if the wrong sensors are deleted.

Then I also add one line to print out the sensors before and after the deleting.

```swift
        print("Sensors:", list.count);
        for (index, sensor) in list.enumerated() {
            print(index, sensor.key, sensor.name,sensor.formattedMiniValue, sensor.formattedValue);
            if let newValue = self.smc.pointee.getValue(sensor.key) {
                // Remove the temperature sensor, if SMC report more that 110 C degree.
                if sensor.type == SensorType.Temperature.rawValue && newValue > 110 {
                    print("Removing",index, sensor.key);
                    list.remove(at: index)
                    continue
                }
                
                if let idx = list.firstIndex(where: { $0.key == sensor.key }) {
                    list[idx].value = newValue
                }
            }
        }
        print("After remove", list.count);
        for(index, sensor) in list.enumerated(){
            print(index, sensor.key, sensor.name,sensor.formattedMiniValue);
            
        }
        self.list = list
```

The results are

```diff
@@Sensors: 32@@
0 Th1H Heatpipe 2 0° 0°C
1 Th2H Heatpipe 3 0° 0°C
2 TC0E CPU diode virtual 0° 0°C
3 TC0F CPU diode filtered 0° 0°C
4 TC0P CPU proximity 0° 0°C
5 TC1C CPU core 2 0° 0°C
6 TC2C CPU core 3 0° 0°C
7 TC3C CPU core 4 0° 0°C
8 TC4C CPU core 5 0° 0°C
9 TC5C CPU core 6 0° 0°C
10 TC6C CPU core 7 0° 0°C
- 11 TC7C CPU core 8 0° 0°C
@@Removing 11 TC7C@@
12 TC8C CPU core 9 0° 0°C
@@Removing 12 TC8C@@
- 13 TCGC GPU Intel Graphics 0° 0°C
14 TG0P GPU proximity 0° 0°C
15 Tm0P Mainboard 0° 0°C
[...]

PR #171, opened 2020-11-18
- fix wrong deletion while enumerating (issue #170)

Military-Doctor:


Military-Doctor:
Fixed issue #170 

Before:
image

After:
image

Issue #169, opened 2020-11-16
RAM panel shows incorrect and duplicate top processes


Yoshi2889:
**Describe the bug**
The RAM panel does not show the most up-to-date processes as shown in the screenshot here (Activity Monitor is sorted by Memory and Safari was not running at the time):

<img width="789" alt="image" src="https://user-images.githubusercontent.com/1280380/99288689-47a3d480-283c-11eb-8ed2-fb838b784c2f.png">

**Details:**
 - Device: 13" Macbook Pro 2017
 - MacOS: 11.0.1
 - Application version: 2.4.3

exelban:
Hi.

What do you mean by `duplicate top processes`?
Does it shows always the Safari on the top of the list? 

Could you please run this command: `/usr/bin/top -l 1 -o mem -n 8 -stats pid,command,mem`?
It will show the top processes.

Yoshi2889:
Thanks for your reply.

As you can see in the screenshot, WindowServer is listed multiple times despite only ever having one process running.

The output is as follows:
<img width="562" alt="image" src="https://user-images.githubusercontent.com/1280380/99290191-3e1b6c00-283e-11eb-8eb8-05a8976e9d9c.png">

This is -again- with Safari completely quit (not closed) for at least 10 seconds. 

While writing this comment, it seems to have gotten even more confused:
<img width="291" alt="image" src="https://user-images.githubusercontent.com/1280380/99290285-61461b80-283e-11eb-95e8-9a5d73978d75.png">

There is only one Safari and WindowServer process:
```
▶ top -l 1 -o mem -stats pid,command,mem | grep -e Safari -e WindowServer
53211  Safari           608M 
125    WindowServer     559M 
53218  com.apple.Safari 15M  
556    com.apple.Safari 12M  
2555   com.apple.Safari 7096K
53230  com.apple.Safari 4996K
53240  com.apple.Safari 3828K
675    SafariBookmarksS 3620K
2600   SafariLaunchAgen 1720K
623    SafariCloudHisto 1608K
2615   com.apple.Safari 1444K
```

exelban:
It looks like everything is ok. The reader just takes the output of this command. And visualize it.

Could you run `top -o mem -stats pid,command,mem`. And check side by side if the top processes in the popup and in the top are the same? It can be some latency because they will have different read moments.

Yoshi2889:
[...]

PR #174, opened 2020-11-19
Partially fix duplicate entries in RAM module

Yoshi2889:
This is related to #169. While this PR does not fix the issue, it does prevent duplicate processes from appearing in the RAM list due to not having enough processes returned from `top`. 

Should this possibly be applied to the other modules as well?

### DIFF
--- a/Modules/Battery/popup.swift
+++ b/Modules/Battery/popup.swift
@@ -255,7 +255,7 @@ internal class Popup: NSView, Popup_p {
     public func processCallback(_ list: [TopProcess]) {
         DispatchQueue.main.async(execute: {
             if !(self.window?.isVisible ?? false) && self.processesInitialized {
-                return;
+                return
             }
             
             if list.count != self.processes.count {
--- a/Modules/Battery/popup.swift
+++ b/Modules/Battery/popup.swift
@@ -260,14 +260,14 @@ internal class Popup: NSView, Popup_p {
             
             if list.count != self.processes.count {
                 self.processes.forEach { processView in
-                    processView.clear();
+                    processView.clear()
                 }
             }
             
             for i in 0..<list.count {
                 let process = list[i]
                 let index = list.count-i-1
-                self.processes[index].attachProcess(process);
+                self.processes[index].attachProcess(process)
                 self.processes[index].value = "\(process.usage)%"
             }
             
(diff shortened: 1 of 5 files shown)


exelban:
Hi.

According to this [issue](https://github.com/exelban/stats/issues/169). This bug (duplicating processes) was fixed. So this fix not necessary. But it could be if you change a little bit the moment when cleaning the process view.

Just check if `self.processes` count and `list` count are different. If yes, clean all views.

Modules/Battery/popup.swift
```diff
@@ -254,20 +254,24 @@ internal class Popup: NSView, Popup_p {
...
-                        self.processes[index].icon = process.icon
-                        self.processes[index].toolTip = "pid: \(process.pid)"
-                    }
+            if !(self.window?.isVisible ?? false) && self.processesInitialized {
+                return;
+            }
+            
+            if list.count != self.processes.count {
+                self.processes.forEach { processView in
+                    processView.clear();
```
exelban, line 263: Remove semicolons please

Yoshi2889:
Thanks for your feedback! I hope I incorporated it properly in my new commits. Please let me know if things need changing.

I took the liberty of doing a little bit of code cleanup, as there is a lot of duplication going on.

exelban:
Thanks for implementing this in all modules which have top processes.

Modules/CPU/popup.swift
```diff
@@ -236,20 +236,24 @@ internal class Popup: NSView, Popup_p {
...
-                    let process = list[i]
-                    let index = list.count-i-1
-                    if self.processes.indices.contains(index) {
-                        self.processes[index].label = process.name != nil ? process.name! : process.command
-                        self.processes[index].value = "\(process.usage)%"
-                        self.processes[index].icon = process.icon
-                        self.processes[index].toolTip = "pid: \(process.pid)"
-                    }
+            if !(self.window?.isVisible ?? false) && self.processesInitialized {
+                return;
```
exelban, line 240: Still a few lefts)

Yoshi2889:
That should be all of them. I need to spend some more time with Swift, haha.

exelban:
Thank you for your contribution. In swift semicolons unnesesary)

PR #198, opened 2020-11-28
Update "Homebrew" installation procedure.

KelSolaar:
`brew cask install` is deprecated:

> Warning: Calling brew cask install is deprecated! Use brew install [--cask] instead.

The *Homebrew* installation procedure thus becomes: `brew install --cask stats`


exelban:
Hi. Thanks.

Issue #197, opened 2020-11-28
Feature Request: Fan speed & temp sensor stat in one icon


yokonrad:
It would be great to have an option to display fan speed and temperature sensor (cpu) stats in one icon similar to smcfancontrol. Of course without control of fans. If they are separated they consume much space which could be saved when merged into the one. Additionaly smcfancontrol is not updated to operate on big sur on 13 inch macbook pro retina. There is an issue at their github but project seems to be dead. Picture attached how it looks with smcfancontrol. I would totally drop of using their app if you could add this function to your app :)

![smcfancontrol-screenshot](https://user-images.githubusercontent.com/25588396/100521092-d3edba00-31a1-11eb-89ef-bb22cb5395ff.png)

exelban:
Hi. Thanks for the feedback.

You can't do it with one icon. But you can combine two: sensors and fans. The difference will be in the layout. 
But it allows you to specify what exactly temperature you want to see. 

<img width="106" alt="Zrzut ekranu 2020-11-28 o 19 09 38" src="https://user-images.githubusercontent.com/13332412/100523011-5334bb00-31ad-11eb-8726-ddffefed9b49.png">

There will be no option to show temperature and fan in one icon. Because it's a separate modules.

yokonrad:
I know that i can combine two separated icons with temperatures and fans. But the macbook pro 13' (late 2013) have only one fan and making icon with one stat looks bad when you compare to the others. Just look on the sshots. There is also smcfancontrol not updated on the sshot about which i mentioned earlier

<img width="350" alt="Zrzut ekranu 2020-11-28 o 20 17 09" src="https://user-images.githubusercontent.com/25588396/100524276-035af180-31b7-11eb-8529-a0ce34bc8315.png">
<img width="489" alt="Zrzut ekranu 2020-11-28 o 20 17 49" src="https://user-images.githubusercontent.com/25588396/100524281-07870f00-31b7-11eb-90df-0d9fe26076e9.png">

exelban:
Ok, I see.
I think I can add fans to the sensors module. And then you could select the temperature sensors and fan in the sensor module.

It will look like this.
<img width="78" alt="Zrzut ekranu 2020-11-28 o 20 50 35" src="https://user-images.githubusercontent.com/13332412/100524796-68b0e180-31bb-11eb-8542-89400146023a.png">

yokonrad:
[...]

PR #200, opened 2020-11-29
Mac Mini detection

luckman212:
seems like b14436b0c3d8cbf73c05cbcbf28c1db3a5f01aa8 broke model detection for me (Mac Mini 2018) and I expect other Mac Mini's (the 2nd "m" in model identifier should be lowercase)

image

image

I created this PR but I couldn't actually test it because I couldn't figure out how to build it, it was complaining about a missing signing cert for the `LaunchAtLogin` module.

### DIFF
--- a/StatsKit/SystemKit.swift
+++ b/StatsKit/SystemKit.swift
@@ -337,11 +337,11 @@ public class SystemKit {
 
 let deviceDict: [String: model_s] = [
     // Mac Mini
-    "MacMini6,1": model_s(name: "Mac mini (Late 2012)", year: 2012, type: .macMini),
-    "MacMini6,2": model_s(name: "Mac mini (Late 2012)", year: 2012, type: .macMini),
-    "MacMini7,1": model_s(name: "Mac mini (Late 2014)", year: 2014, type: .macMini),
-    "MacMini8,1": model_s(name: "Mac mini (Late 2018)", year: 2018, type: .macMini),
-    "MacMini9,1": model_s(name: "Mac mini (M1, 2020)", year: 2020, type: .macMini),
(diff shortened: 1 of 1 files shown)


exelban:
Hi. Thanks for this. I just look at other models and think that makes mistakes in mac mini models.
You need to select your own certificate to build the app. But still, it looks like everything is ok. Thanks.

luckman212:
@exelban Thank you. I did select my certificate but it looks like maybe there is somewhere else within the project that has a hardcoded dev ID or something, because it still won't build.

luckman212:
**update:** I figured out how to make it build. Had to go one by one through each Target and update the codesigning cert. I tried to find a way to bulk-update the settings for the entire project but it doesn't seem possible...

Anyway, new build tested & working... 🎉 

image

macmeister:
Fantastic tool. Thank you for all the hard work! This is not a priority or important to me, but since this tool seems just about perfect, I thought you might want to know that I too see a variance in the CPU detected on the latest 2.4.6 (and earlier).
And FYI from System Info:
  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro11,3
  Processor Name:	Quad-Core Intel Core i7
  Processor Speed:	2.3 GHz

Image 2020-11-30 at 10 12 06 AM

exelban:
@macmeister thanks. I know about this, but have no idea how to detect the mac better. Please take a look [here](https://github.com/exelban/stats/issues/197#issuecomment-735650380_) why it's happend.

macmeister:
Interesting. Thanks for sharing! Not sure why I didn't discover your "stats" earlier. Been using MenuMeters for years. Stats does use .2% more CPU, but well worth it with the extra details and sexiness! ;) Have a blessed week.

PR #229, opened 2020-12-16
Update OS detection

qkdxorjs1002:
New version numbering rules

### DIFF
--- a/StatsKit/SystemKit.swift
+++ b/StatsKit/SystemKit.swift
@@ -407,4 +407,5 @@ let osDict: [String: String] = [
     "10.14": "Mojave",
     "10.15": "Catalina",
     "11.0": "Big Sur",
+    "11.1": "Big Sur"
 ]
--- a/StatsKit/SystemKit.swift
+++ b/StatsKit/SystemKit.swift
@@ -101,8 +101,7 @@ public class SystemKit {
             build = buildArr[1].replacingOccurrences(of: "Build ", with: "").replacingOccurrences(of: ")", with: "")
         }
         
-        let version = "\(systemVersion.majorVersion)" +
-            (systemVersion.majorVersion >= 11 ? "" : ".\(systemVersion.minorVersion)")
+        let version = "\(systemVersion.majorVersion)" + (systemVersion.majorVersion >= 11 ? "" : ".\(systemVersion.minorVersion)")
         self.device.os = os_s(name: osDict[version] ?? LocalizedString("Unknown"), version: systemVersion, build: build)
         
         self.device.info?.cpu = self.getCPUInfo()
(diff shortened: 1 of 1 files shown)


exelban:
Hi.

Thanks for your contribution. Does it fix something?

qkdxorjs1002:
> Hi.
> 
> Thanks for your contribution. Does it fix something?

This commit includes support (fix unknown version) for version 11.1 (Big sur) and later with new version numbering rule.
ex) 12.0.1 (Codename), 12.1 (Codename), 13.0.1 (Codename2)

exelban:
As far as I can see, it does not solve anything. The `11.0` was already on the list. So it does not fix the unknown version for Big Sur.
How it fix the next version's detection?

exelban:
It just provides 2 dimension array. And create a potential bug. Because in line 106. You suppose the major version always be on the list. And if apple tomorrow releases a new version 12.0. The app will just crash(

qkdxorjs1002:
Thanks for your comment. I am new to Swift.
I've used optional question marks to deal with nil of nested dict.

And this is intended to improve what you have to react to every time for minor patches.

exelban:
No problem. 
These changes do not fix anything. And do not help with new versions.

Maybe will be better if you describe what do you want to achieve with these changes?

exelban:
Also, now I cannot merge this at all. Because it will not work properly with the next versions of macOS.

`11.x` - it's not a macOS Big Sur.
`11.0` - it's a macOS Big Sur.

qkdxorjs1002:
I've just intended to distinguish codename without condition of minor version on major version 11 and over.

exelban:
But `11.1` does not have a name for now. We will know it only at WWDC 2021.
And are you sure that the next version will be `11.1` not `11.10`?

exelban:
Take a look at [this](https://en.wikipedia.org/wiki/MacOS_version_history).

qkdxorjs1002:
umm...can I ask my macos version is wrong...?


스크린샷 2020-12-20 오전 1 43 17

exelban:
Wow. I suppose Apple lose in numbers)

Zrzut ekranu 2020-12-19 o 17 49 18

exelban:
I saw this a few times with iOS, but never with macOS)

qkdxorjs1002:
Oh, I'm sorry for not being able to communicate clearly.
I thought you knew the 11.1 update was released. hahaha

exelban:
Nope(

exelban:
Now I check and have an update to `11.1.1`. It looks like they release at the beginning the Big Sur under `11.0` And after change their mind) And jump to `11.1`)

Thanks. Now I see what it fixing.

qkdxorjs1002:
It's my pleasure. Sorry to take your time. I will try to clarify the reasons and intentions. :)

exelban:
No problem. Thanks for your contribution.

exelban:
Apple gives names for each minor version. Not major.

PR #259, opened 2020-12-29
Update zh-Hans Translation and Fix selectedInterface

Tai-Zhou:
The words shown in MenuItem now can be localized. I also changed the name of fans which was like "Fan #0". Sorry for the inconvenience that my editor deleted spaces at the end of each line automatically.

### DIFF
--- a/Modules/Battery/settings.swift
+++ b/Modules/Battery/settings.swift
@@ -23,7 +23,7 @@ internal class Settings: NSView, Settings_v {
     private var button: NSPopUpButton?
     
     private var numberOfProcesses: Int = 8
-    private let levelsList: [String] = [LocalizedString("Disabled"), "0.03", "0.05", "0.1", "0.15", "0.2", "0.25", "0.3", "0.4", "0.5"]
+    private let levelsList: [String] = ["Disabled", "0.03", "0.05", "0.1", "0.15", "0.2", "0.25", "0.3", "0.4", "0.5"]
     private var lowLevelNotification: String {
         get {
             return self.store.pointee.string(key: "\(self.title)_lowLevelNotification", defaultValue: "0.15")
@@ -75,7 +75,7 @@ internal class Settings: NSView, Settings_v {
             title: LocalizedString("Low level notification"),
             action: #selector(changeUpdateInterval),
             items: levels,
-            selected: self.lowLevelNotification == "Disabled" ? LocalizedString("Disabled") : "\(Int((Double(self.lowLevelNotification) ?? 0)*100))%"
+            selected: self.lowLevelNotification == "Disabled" ? self.lowLevelNotification : "\(Int((Double(self.lowLevelNotification) ?? 0)*100))%"
         ))
         
         self.addSubview(SelectTitleRow(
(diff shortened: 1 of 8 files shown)


exelban:
Please remove all redundant +- empty lines. I don't see what you try to change.

exelban:
Also, I noticed that you try to just translate values like `At start`. It cannot be done like this. These 2 select options (Update intervals and colors) not prepared for translations. There must be 2 separate values: one for key and another to display. I haven't done it yet.

And at this stage, you can commit the translations. But they will be not visible now.

Tai-Zhou:
> Also, I noticed that you try to just translate values like `At start`. It cannot be done like this. These 2 select options (Update intervals and colors) not prepared for translations. There must be 2 separate values: one for key and another to display. I haven't done it yet.

Sorry for bothering, I've seen my mistakes. The spaces were added back and some changes reverted. The modification is now limited to the "Disabled" of the battery notification, the "Autodetection" of network interface and the name of fans.

Modules/Battery/settings.swift
```diff
@@ -110,7 +110,7 @@ internal class Settings: NSView, Settings_v {
     }
     
     @objc private func changeUpdateInterval(_ sender: NSMenuItem) {
-        if sender.title == "Disabled" {
+        if sender.title == LocalizedString("Disabled") {
```
exelban, line 113: Sorry, but the saving key cannot be translated. It must be constant.

Modules/Net/settings.swift
```diff
@@ -123,7 +123,7 @@ internal class Settings: NSView, Settings_v {
     @objc func handleSelection(_ sender: NSPopUpButton) {
         guard let item = sender.selectedItem else { return }
         
-        if item.title == "Autodetection" {
+        if item.title == LocalizedString("Autodetection") {
```
exelban, line 126: The same problem as in the Battery module.

Modules/Battery/settings.swift
Tai-Zhou: Thank you for your attention! I just made a new commit to keep the `lowLevelNotification` and saving term `Battery_lowLevelNotification` with a constant value "Disabled".

Modules/Net/settings.swift
Tai-Zhou: https://github.com/exelban/stats/blob/e33082f0cab3b2ea171c4cb06848c459843f6897/Modules/Net/settings.swift#L126-L127
When `item.title == "Autodetection"`, the term `Network_interface` would just be removed, so I think saving state would not be changed.

exelban:
Hi. 
I still see that you try to translate not prepared values (`LocalizedString("Disabled")` and `LocalizedString("Autodetection") `). I will not merge it. See my second post why.

Tai-Zhou:
Happy New Year! 🎇🎆
I just realized I've misunderstood your meaning of "key", I thought it was the key of storage 😅
https://github.com/exelban/stats/blob/6fab32188443e7e22e3f490a15e525a399e5ce36/Modules/Net/settings.swift#L111-L115
I made a new commit for `LocalizedString("Autodetection")`. Now both the selection and storage don't depend on the localization anymore. So I think defining a `[KeyValue_t]` would be a little bit redundant for this part. I also find the line 115 in the code block above should be moved before `selectItem` to make it work properly.
For `LocalizedString("Disable")`, it is indeed not a good modification, I can revert this part.

exelban:
Hi. Sorry for the long response.

Please do not translate the values which not ready for translation. If you want to have these values translated you can help to make it ready. You can take a look [here](https://github.com/exelban/stats/blob/f17f38f0fe850eaeb38e5a1cb2cce981536aa4ce/StatsKit/helpers.swift#L38-L43) how I make select option translatable. It must have a key, which will be used to store. And value (text) which could be translated.

Tai-Zhou:
I see, that's very kind of you with great patience, and sorry for troubling you for such a long time. I've reverted all the changes except the `selectedInterface`, and a new version of zh-Hans translation is updated.

PR #261, opened 2020-12-30
Added pt-PT localization

AdamModus:
I'm a big fan of this app, been using it for months! Thought I'd give a little bit in return!

Also, in case some people aren't aware, Portuguese from Portugal is not the exact same as the Portuguese in Brazil. They are mutually intelligible but have much bigger differences between them than, for example, en-US and en-UK.


exelban:
Can I add you as a contributor to this translation?

AdamModus:
Hi there. Sure, feel free to do that 😄

PR #275, opened 2021-01-04
Single click to download latest dmg

chetanddesai:
Minor usability improvement, save people an extra click 😉

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ macOS system monitor in your menu bar
 
 ## Installation
 ### Manually
-You can download the latest version [here](https://github.com/exelban/stats/releases).  
-Download the file called `Stats.dmg`. Open it and move the app to the application folder.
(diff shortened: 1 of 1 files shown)


Military-Doctor (commented):
Hi, thanks for the contribution!

Both the original version and your version is good.

In your version, user can directly download the latest version of Stats.dmg.
However, the original version can let user know the newest feature and bug-fixes.

I think it is better of the original version, what do you think?

(by the way, I'm not the author...

exelban:
Thanks for your contribution.

chetanddesai:
> Hi, thanks for the contribution!
> 
> Both the original version and your version is good.
> 
> In your version, user can directly download the latest version of Stats.dmg.
> However, the original version can let user know the newest feature and bug-fixes.
> 
> I think it is better of the original version, what do you think?
> 
> (by the way, I'm not the author...

I could certainly add a sentence or something that says to view the latest feature and bug-fixes in the release notes to check out the [releases](https://github.com/exelban/stats/releases) page.  Let me know!

Issue #313, opened 2021-01-23
Feature Request: Smooth Chart


Military-Doctor:
Hi, Sorry to bother you, it's me again )

If the CPU usage history chart can be smooth, then statistics should have a better look!

Could we have this feature?

I started a new branch and did some work on it.

Left: Smoothed
Right: Original

![Xnip2021-01-23_22-24-38](https://user-images.githubusercontent.com/14367694/105584120-4c9c7b80-5dcb-11eb-99ec-26e28b0bb261.png)

![Xnip2021-01-23_22-34-45](https://user-images.githubusercontent.com/14367694/105584217-532af300-5dcb-11eb-8ae3-71d359743ee7.png)

Military-Doctor:
![image](https://user-images.githubusercontent.com/14367694/105606424-5d9dba80-5dd4-11eb-9743-96064378bcd7.png)

exelban:
Hi. Thanks. I'm not sure if it's better. Because a lot of information is missing when smoothing the chart.

Military-Doctor:
Maybe we could give users an option to enable smooth charts?

Military-Doctor:
Hi, thanks for replying.

What do you think of this? The user can choose whether to open the smooth chart in settings page. 
After all, some users may prefer visual smoothness. 😺 

![image](https://user-images.githubusercontent.com/14367694/105609111-fa1a8980-5de1-11eb-9205-e74a7163bddf.png)

I have tested this feature and am ready to submit to the master branch.

exelban:
Hi. Thanks for the proposal. I'm 99% sure that this is a redundant option. The chart must show the usage. And if the chart will be smoothed, the information about usage in short periods of time will be loosed. 

Also, it cannot be just a toggle. A future like this must be realized as a slider. Where you can specify the level of smoothness for each chart individually.

exelban:
Please reopen this issue, if you still want to add this feature to the app.

Military-Doctor:
I am sorry for the late reply. Sincerely speaking that I think you're right, this is a redundant feature. I will reopen this issue if someone needed this feature later. 😃

PR #314, opened 2021-01-23
feat: add smooth chart support.

Military-Doctor:
This pull request includes 1 new feature: smooth chart.

Note: network chart are not affected by this update.

**Preview:**
Left: smooth enabled
Right: original


exelban:
Hi. I will not merge this feature. The reason I describe in the #313. I suppose it could be merged if you will add some smooth range. But not like the toggle button.

PR #337, opened 2021-02-05
Update Chinese Simplified translation

Tai-Zhou:
Fix confusing translation.


exelban:
Hi. Thanks.

Could I ask you to make a more readable PR name? Something like this: `Update Chinese Simplified translation`. It will be easier for me when I making a change list. Because I don't know what `zh-Hans` means.

Tai-Zhou:
> Could I ask you to make a more readable PR name? Something like this: `Update Chinese Simplified translation`. It will be easier for me when I making a change list. Because I don't know what `zh-Hans` means.

Sorry I just followed the directory name, changed to the friendly title now. 😵

PR #338, opened 2021-02-06
Update macOS Name Recognition

Tai-Zhou:
macOS 11.2 was released. It seems Apple adjusted the naming of macOS version numbers to a way similar to the iOS/iPadOS, so I guess only the `majorVersion` is needed in the future.

### DIFF
--- a/StatsKit/SystemKit.swift
+++ b/StatsKit/SystemKit.swift
@@ -115,7 +115,7 @@ public class SystemKit {
             build = buildArr[1].replacingOccurrences(of: "Build ", with: "").replacingOccurrences(of: ")", with: "")
         }
         
-        let version = "\(systemVersion.majorVersion).\(systemVersion.minorVersion)"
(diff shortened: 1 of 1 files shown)


qkdxorjs1002:
exelban didn't accept this method of recognition... #229

Tai-Zhou:
> exelban didn't accept this method of recognition... #229

Wow, sorry for missing your PR. I just wanted to add `"11.2": "Big Sur"` at the beginning, and thought it annoying to change per season and made this modification.

Maybe we will know what Apple is thinking until the release of next major version at WWDC2021.

qkdxorjs1002:
> > exelban didn't accept this method of recognition... #229
> 
> Wow, sorry for missing your PR. I just wanted to add `"11.2": "Big Sur"` at the beginning, and thought it annoying to change per season and made this modification.
> 
> Maybe we will know what Apple is thinking until the release of next major version at WWDC2021.

I think so, too. )

exelban:
Hi. Thanks for the PR.

@qkdxorjs1002 Yes. It's the same changes as you propose in #229. But now it more clear, that it was not a mistake with `11.0` and `11.1`. But Apple moved from a minor version to a major with macOS.

So now, these changes make sense.

PR #340, opened 2021-02-08
Add Localization to the Power Source

Tai-Zhou:
Add two terms to all the languages and add localizations of widgets to Bulgarian and Romanian.

Before localization: before

After localization: after

### DIFF
--- a/Modules/Battery/popup.swift
+++ b/Modules/Battery/popup.swift
@@ -224,7 +224,7 @@ internal class Popup: NSView, Popup_p {
             self.dashboardBatteryView?.setValue(abs(value.level))
             
             self.levelField?.stringValue = "\(Int(abs(value.level) * 100)) %"
-            self.sourceField?.stringValue = value.powerSource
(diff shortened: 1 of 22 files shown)


exelban:
Hi. Thanks.

Modules/Battery/popup.swift
```diff
@@ -224,7 +224,7 @@ internal class Popup: NSView, Popup_p {
             self.dashboardBatteryView?.setValue(abs(value.level))
             
             self.levelField?.stringValue = "\(Int(abs(value.level) * 100)) %"
-            self.sourceField?.stringValue = value.powerSource
+            self.sourceField?.stringValue = "\(LocalizedString(value.powerSource))"
```
exelban, line 227: I noticed it after merge. `LocalizedString` returns a String. So you don't need to pass it throw quotes. It could be simplified with this: `self.sourceField?.stringValue = LocalizedString(value.powerSource)`.

But it's ok)

PR #341, opened 2021-02-08
Update 🇹🇼Taiwan Traditional Chinese Translation File

jrthsr700tmax:
Update 🇹🇼Taiwan Traditional  Chinese Translation File
更新🇹🇼臺灣正體中文翻譯檔案

*Update Charts Name in Widgets Item (更新「小工具」項下之圖表名稱)

### DIFF
--- a/Stats/Supporting Files/zh-Hant.lproj/Localizable.strings
+++ b/Stats/Supporting Files/zh-Hant.lproj/Localizable.strings
@@ -82,15 +82,15 @@
 "Display mode" = "顯示模式";
 "One row" = "單行";
 "Two rows" = "雙行";
-"Mini widget" = "Mini";
-"Line chart widget" = "Line chart";
-"Bar chart widget" = "Bar chart";
-"Pie chart widget" = "Pie chart";
-"Network chart widget" = "Network chart";
-"Speed widget" = "Speed";
-"Battery widget" = "Battery";
-"Text widget" = "Text";
-"Memory widget" = "Memory";
+"Mini widget" = "小型";
+"Line chart widget" = "現狀圖";
+"Bar chart widget" = "條狀圖";
+"Pie chart widget" = "圓形圖";
+"Network chart widget" = "網狀圖";
+"Speed widget" = "速度";
+"Battery widget" = "電池";
+"Text widget" = "文字";
+"Memory widget" = "記憶體";
 
 // Module Kit
 "Open module settings" = "打開模組設定";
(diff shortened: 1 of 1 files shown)


exelban:
Hi. Thanks.

Could I ask you to make a more readable PR name? Something like this: `Update Chinese Translation translation`. It will be easier for me to create a change list.

I'm not good at Chinese. What is the difference between `Taiwan Traditional Chinese` and `Chinese Traditional`?
XCode says that this is a translation for `Chinese Traditional` (zh-Hant). But I also see the `Chinese Traditional, Taiwan` in the list of available languages in the XCode.

Maybe you want to create a new translation? Or it's the same?

jrthsr700tmax:
Thanks to your feedback.

There are two major divisions in Traditional Chinese, one is Taiwan and another is Hong Kong actually. Because of I’m Taiwanese, that I use the Taiwan one for the major Traditional Chinese translation and localization in this App. Because of the character count limitations of the title, it is the reason why I used “Taiwan Trad. Chinese” in the title, but I’ve fix it back to “Taiwan Traditional Chinese.”

By the way, because of I’m Taiwanese, and many terms are different between Taiwan and Hong Kong, also I only can translate with Taiwan’s terms. If the team has a request for Hong Kong Traditional Chinese translation, you can ask someone live in Hong Kong or Hong Kong People to make another locale file for Hong Kong’s localization. And I’m very glad to see that.

Sincerely,
Your best 

jrthsr700tmax

exelban:
Thanks for the explanation)

I think we could leave it for now. If someone will want to help with Hong Kong Traditional Chinese. We will just branch a new translation for it. Till no one say that something is wrong with `Traditional Chinese` it could stay like this.

jrthsr700tmax:
Thank you!

—————————

in Chinese 

感謝！

Stats/Supporting Files/zh-Hant.lproj/Localizable.strings
```diff
@@ -82,15 +82,15 @@
...
-"Line chart widget" = "Line chart";
-"Bar chart widget" = "Bar chart";
-"Pie chart widget" = "Pie chart";
-"Network chart widget" = "Network chart";
-"Speed widget" = "Speed";
-"Battery widget" = "Battery";
-"Text widget" = "Text";
-"Memory widget" = "Memory";
+"Mini widget" = "小型";
+"Line chart widget" = "現狀圖";
```
Tai-Zhou, line 86: This should be "折線圖" (or "線狀圖" according to the pronunciation).

jrthsr700tmax: Thanks for your comment.

I’ve fix the translation of “Line chart widget” form the wrong one “現狀圖” to the correct one “線狀圖.”

Sincerely,
Your best

jrthsr700tmax

—————————
in Chinese 

感謝您的留言。

本人業已修正「Line chart widget」之翻譯，業將錯誤翻譯「現狀圖」修正為正確翻譯「線狀圖」。

敬祝
順頌時祺

jrthsr700tmax

PR #343, opened 2021-02-11
Update Polish translation file

yokonrad:
Hello. I changed a few lines to proper polish translation. Not sure about GPU "Utilization", because i have MBPr 13' with Iris and GPU widget is not showing this translation, but i think it should be good.


exelban:
Hi. Thanks.

Tai-Zhou:
> Hello. I changed a few lines to proper polish translation. Not sure about GPU "Utilization", because i have MBPr 13' with Iris and GPU widget is not showing this translation, but i think it should be good.

You can find "Utilization" at GPU popup after click the "Details" button, the value is the same as "GPU Utilization".

Utilization

PR #367, opened 2021-02-26
Match ip addr

### LINKED ISSUE #366: Public IP shown as "<!DOCTYPE html>"
**Describe the bug**
See the sceenshot, looks like a content of a html document is shown instead of public IP

Screenshot 2021-02-26 at 19 09 20

**Details:**
 - MacOS 11.2.1
 - Application version 2.5.1

tlove004:
fixes #366

### DIFF
--- a/Modules/Net/popup.swift
+++ b/Modules/Net/popup.swift
@@ -326,8 +326,10 @@ internal class Popup: NSView, Popup_p {
                     self.ssidField?.stringValue = LocalizedString("Unavailable")
                 }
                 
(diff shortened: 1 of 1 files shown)


exelban:
Hi. Thanks for the PR. 
There is a little bit better way to fix it. I take a look at the function fetching public IP. And there is no check if the response status is 200. And I suppose DOCTYPE appears because of non 200 status. And this bug #366 must be fixed in this way.

tlove004:
Fair enough -- this was an admittedly very quick fix.

PR #382, opened 2021-03-16
Update Polish translation file

yokonrad:
Hi, small fixes in Polish translation file

### DIFF
--- a/Stats/Supporting Files/pl.lproj/Localizable.strings
+++ b/Stats/Supporting Files/pl.lproj/Localizable.strings
@@ -2,7 +2,7 @@
 //  Localizable.strings
 //  Stats
 //
-//  Created by Yoko on 15/03/2021.
+//  Created by Serhiy Mytrovtsiy on 27/08/2020.
 //  Using Swift 5.0.
 //  Running on macOS 10.15.
 //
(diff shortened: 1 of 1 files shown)


exelban:
Hi. Thanks.

Stats/Supporting Files/pl.lproj/Localizable.strings
```diff
@@ -2,7 +2,7 @@
 //  Localizable.strings
 //  Stats
 //
-//  Created by Serhiy Mytrovtsiy on 27/08/2020.
+//  Created by Yoko on 15/03/2021.
```
exelban, line 5: Why did you change this?

yokonrad: Because i saw in other language files that the authors of the modified versions have changed this line. You can skip this line if you don't want to change it.

exelban: I'm ok with update credentials if you are an author of translation or contribute a lot to the existing one.

yokonrad: Well, im using your app and it wont change for a long time :) I used menumeters for 6-7 years before. So i will try to update it every time when it will be needed to.

exelban: Thanks. I'm trying to keep Polish translation up to date (it's my primary language on the main machine))

yokonrad: Ah, okey. Didnt know :) Then just skip those lines.

PR #393, opened 2021-03-31
Fixed spanish lang, network refresh button width fix

ppizarror:
- Added missing refresh lang entry to langs.
- Refresh button now computes the width dynamically to support new languages. The fixed 44px creates visual glitches when using my Spanish translated text.
- Updated spanish localization.
- Fixed english update lang.
- Fixed vertical centering of update message, now it works with other languages (Spanish).

### DIFF
--- a/Modules/Net/popup.swift
+++ b/Modules/Net/popup.swift
@@ -185,8 +185,10 @@ internal class Popup: NSStackView, Popup_p {
         
         let row: NSView = NSView(frame: NSRect(x: 0, y: 0, width: view.frame.width, height: Constants.Popup.separatorHeight))
         
+        let updatebutton_width = (LocalizedString("Refresh") as NSString).size(withAttributes: [NSAttributedString.Key.font: NSFont.systemFont(ofSize: 11, weight: .light)]).width
+        
         let button = NSButtonWithPadding()
-        button.frame = CGRect(x: view.frame.width - 44, y: 5, width: 44, height: 20)
+        button.frame = CGRect(x: view.frame.width - updatebutton_width, y: 5, width: updatebutton_width, height: 20)
         button.bezelStyle = .regularSquare
         button.isBordered = false
         button.action = #selector(self.refreshPublicIP)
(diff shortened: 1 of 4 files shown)


exelban:
The `Refresh` button it's a temporary solution. And I wasn't sure if it must be translated. Because I will redraw the whole popup view. And I want to use icon, not text for this action. But thanks.

Modules/Net/popup.swift
```diff
@@ -185,8 +185,10 @@ internal class Popup: NSStackView, Popup_p {
         
         let row: NSView = NSView(frame: NSRect(x: 0, y: 0, width: view.frame.width, height: Constants.Popup.separatorHeight))
         
+        let updatebutton_width = (LocalizedString("Refresh") as NSString).size(withAttributes: [NSAttributedString.Key.font: NSFont.systemFont(ofSize: 11, weight: .light)]).width
+        
         let button = NSButtonWithPadding()
-        button.frame = CGRect(x: view.frame.width - 44, y: 5, width: 44, height: 20)
+        button.frame = CGRect(x: view.frame.width - updatebutton_width, y: 5, width: updatebutton_width, height: 20)
```
exelban, line 191: There is a little bit better way to do it.

First move `LocalizedString("Refresh")` to separate variable. Then you can use it in 3 places: for width, in the button title, and toolTip. I suppose the font for button text could be moved to a separate variable too.
Also, there is no reason for conversion to the `NSString`. `LocalizedString` returns `String` which confirms `NSString`.

PS: the code base of Stats use camel case (at least for variables)

ppizarror: Thanks, I'll change this ASAP. Also think that an icon is a better approach

ppizarror:
I've added a new reloading icon, let me know if it looks nice to you for adding to this PR

image

ppizarror:
@exelban what was the reason for closing this PR without merging the spanish lang? There was an issue with the translation?

ppizarror:
I can create only a spanish .strings if I'm not allowed to touch the code

exelban:
Hi. Oh, sorry. I close it because think that it has the same #395.
Nope, I will merge this one. My bad(

ppizarror:
Yes, I'll create a new PR with epic icons for the reload button, but I'm struggling to keep my MSc. and the work and family, also covid does not make it easier. Thanks for your hard work, love from Chile!!

exelban:
Thanks for your contribution. And sorry for my mistake with PR.

PR #411, opened 2021-04-09
Add Missing Localizations

Tai-Zhou:
Add missing localizations for all languages to keep the files with the same line numbers. Chinese Simplified translation is also updated.

### DIFF
--- a/Stats/Supporting Files/bg.lproj/Localizable.strings
+++ b/Stats/Supporting Files/bg.lproj/Localizable.strings
@@ -24,6 +24,7 @@
 "Yes" = "Да";
 "No" = "Не";
 "Automatic" = "Автоматично";
+"Manual" = "Manual";
 "None" = "Без";
 "Dots" = "Точки";
 "Arrows" = "Стрелки";
@@ -31,6 +32,8 @@
 "Short" = "Кратко";
 "Long" = "Дълго";
 "Statistics" = "Статистика";
+"Max" = "Max";
+"Min" = "Min";
 
 // Alerts
 "New version available" = "Налична е нова версия";
@@ -95,6 +98,7 @@
 // Module Kit
 "Open module settings" = "Отваряне на настройките на модула";
 "Select widget" = "Избор на уиджет %0";
+"Open widget settings" = "Open widget settings";
 "Update interval" = "Интервал на опресняване";
 "Usage history" = "История на използването";
 "Details" = "Подробности";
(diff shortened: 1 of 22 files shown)


Tai-Zhou:
I write a Python script to sync translations across different languages. This script relies on the correct sequence of strings, so I made a commit (https://github.com/exelban/stats/pull/411/commits/f271addfa1d5ad8c32770a98befeb42bee85af20) to fix the Japanese L10n file.
```python
import os

L10nPath = "Stats/Supporting Files/"
langList = os.listdir(L10nPath)
dic = set()
EngL10n = open(L10nPath + "en.lproj/Localizable.strings").readlines()
for i in range(11, len(EngL10n)):
    if EngL10n[i][0] != '/' and EngL10n[i][0] != '\n':
        dic.add(EngL10n[i].split(" = ")[0])

# Add English Missing L10n
for lang in langList:
    if lang[-6:] == ".lproj" and lang != "en.lproj":
        langL10n = open(L10nPath + lang + "/Localizable.strings").readlines()
        i = 12
        while i < len(langL10n):
            term = langL10n[i].split(" = ")[0]
            if term != EngL10n[i].split(" = ")[0]:
                if langL10n[i][0] != '/' and langL10n[i][0] != '\n' and term not in dic:
                    dic.add(term)
                    EngL10n.insert(i, term + " = " + term + ";\n")
                else:
                    langL10n.insert(i, EngL10n[i])  # Make line number the same as English
            i += 1
EngL10nFile = open(L10nPath + "en.lproj/Localizable.strings", 'w')
for line in EngL10n:
    EngL10nFile.write(line)
EngL10nFile.close()

# Add Other Languages' Missing L10n
for lang in langList:
    if lang[-6:] == ".lproj" and lang != "en.lproj":
        langL10n = open(L10nPath + lang + "/Localizable.strings").readlines()
        for i in range(11, len(EngL10n)):
            if EngL10n[i][0] != '/' and EngL10n[i][0] != '\n' and EngL10n[i].split(" = ")[0] != langL10n[i].split(" = ")[0]:
                langL10n.insert(i, EngL10n[i])
        langL10nFile = open(L10nPath + lang + "/Localizable.strings", 'w')
        for line in langL10n:
            langL10nFile.write(line)
        langL10nFile.close()
```

exelban:
Hi. It's a nice script. I was thinking about gh actions. Maybe it could be some part of the check for PRs. What do you think?

Tai-Zhou:
> Hi. It's a nice script. I was thinking about gh actions. Maybe it could be some part of the check for PRs. What do you think?

I think GH action is a good idea. How about using the GH action to run the script after PR is merged, then commit and push (maybe like https://github.com/marketplace/actions/github-commit-push).

I'm also concerning that this script is too fragile to run remotely, if the sequence of the translation is changed then it would exit with error. A better solution might be that ask the contributors to sync their translations with the English version (especially adding new strings) and update both their language and English in the same PR, and GH actions can be used to check the PRs. Then after the PR is merged, another GH action can use a simplified script to add new the new string to other languages. I think in this case, the script can be mush more stable.

PR #437, opened 2021-04-25
Update Chinese Simplified

Tai-Zhou:
Translate colors, and try the GH action.


exelban:
Hi. Thanks. I change the action policy. And now it must run on PR.
I take your idea of the script and make a new one. Which just checking the keys in the translation files with English (original).
And it throwing the error if some key does not match (line number).

Issue #444, opened 2021-04-28
Feature Request: country flag icon


altrdev:
A cool feature in iStatMenu is country flag icon to network module.

Example: 
<img width="249" alt="Screenshot 2021-04-28 at 21 13 55" src="https://user-images.githubusercontent.com/22495387/116459944-bc0c8600-a866-11eb-9867-00bf11e8e21c.png">

What do you think about add this feature in stats?

exelban:
Hi. Thanks for the proposal. Now there is a country code near IP address (if it's available). I'm to have a flag against the code. But I will not have time to implement this. Maybe someone will want to contribute. And it could be a nice feature.

exelban:
Hi. 
Could I ask why you close this issue?

PR #455, opened 2021-05-04
[feat]: added country flag - Net PopUp Module

altrdev:
#444 Country flag icon near public ip

### DIFF
--- a/Modules/Net/popup.swift
+++ b/Modules/Net/popup.swift
@@ -260,6 +260,15 @@ internal class Popup: NSStackView, Popup_p {
         })
     }
     
+    private func flag(country:String) -> String {
+        let base : UInt32 = 127397
+        var s = ""
+        for v in country.unicodeScalars {
+            s.unicodeScalars.append(UnicodeScalar(base + v.value)!)
+        }
+        return String(s)
+    }
+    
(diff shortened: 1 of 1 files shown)


Modules/Net/popup.swift
```diff
@@ -260,6 +260,15 @@ internal class Popup: NSStackView, Popup_p {
         })
     }
     
+    private func flag(country:String) -> String {
+        let base : UInt32 = 127397
+        var s = ""
+        for v in country.unicodeScalars {
+            s.unicodeScalars.append(UnicodeScalar(base + v.value)!)
```
exelban, line 267: Hi. Are you sure that `UnicodeScalar` cannot be nil?

exelban:
Hi. I'm going to close the issue at Sanday if there will be no changes.

Modules/Net/popup.swift
altrdev: Cannot be nil. But can be no flag symbol. Closed

exelban: [UnicodeScalar](https://developer.apple.com/documentation/swift/unicode/scalar/2905753-init):
`If v is an invalid Unicode scalar value, the result is nil.`

PR #524, opened 2021-07-04
Update German translation

milotype:


meiser79:
There's a 's' missing in 'Ausrichtung'.

milotype:
@meiser79 Thank you for catching my mistakes. I'm sorry about that.

I've updated the German (and Croatian) in #538.

PR #545, opened 2021-07-15
updated Hungarian translations

majonez:


Stats/Supporting Files/hu.lproj/Localizable.strings
```diff
@@ -3,6 +3,8 @@
 //  Stats
 //
 //  Created by @moriczr on 22/01/2021.
+//  Updated by András Oravecz (info@oandras.hu) on 15/07/2021
+//
```
exelban, line 7: Hi. Please remove these lines. If you want to add your credentials, you can do it in the line above.

majonez: Hi, I changed those lines.

Issue #573, opened 2021-08-03
Feature request: Remember manual fan setting


Atemu:
Currently, when you switch off manual mode and back, the previous manual settings are lost and not re-applied. https://github.com/exelban/stats/issues/570

It'd be very useful if stats could remember the manually set frequencies and re-apply them whenever manual mode is activated.

A related issue is that manually set fan speeds seem to be lost when the machine sleeps; that should be fixed too for this feature to make sense.

exelban:
You could ask in the previous ticket if it will be changed. Because the answer is no. There will be no memory for fan speed. To prevent a situation when you yesterday set the fan speed to max. And today when you switching back from auto to manual mode it spins your fan to max.

exelban:
Maybe it could be realized as an option in the settings. Which is disabled by default.

PR #592, opened 2021-08-22
Update german translation

Fighter456:
Localizes the newly added "Save the fan speed" option for german users.

See #573


exelban:

Atemu (approved):
Translation sounds good to me (German).

PR #627, opened 2021-09-15
Update Italian localization

kevinm6:


Angelk90:
Hi @kevinm6 , I don't know if it's a problem but I get the following result when the battery is fully charged: 
Schermata 2021-09-20 alle 21 33 27

https://github.com/exelban/stats/blob/master/Stats/Supporting%20Files/it.lproj/Localizable.strings#L236

kevinm6:
> Hi @kevinm6 , I don't know if it's a problem but I get the following result when the battery is fully charged: 
> 
> Schermata 2021-09-20 alle 21 33 27
> 
> 
> 
> https://github.com/exelban/stats/blob/master/Stats/Supporting%20Files/it.lproj/Localizable.strings#L236


Thank you @Angelk90 for show me the issue, I'll fix it soon with just "Carica".

Angelk90:
@kevinm6 : I guess it cuts the text, right?

kevinm6:
Yes exactly. 
I’m doing it right now, the we just have to wait @exelban to merge the fix.

Angelk90:
@kevinm6 : 
I also point out this:
https://github.com/exelban/stats/blob/master/Stats/Supporting%20Files/it.lproj/Localizable.strings#L119

Then I noticed that texts like these are not translated: 

Disabled:
Schermata 2021-09-20 alle 22 12 14

Fan #Number:
Schermata 2021-09-20 alle 22 16 48

PR #689, opened 2021-11-07
Fix apple m1 fan.

sdkdimon:
keyInfo_t dataSize must be 32 instead of 64 bytes.


exelban:
Hi. Does it add fans for apple silicon?

sdkdimon:
Yes.

exelban:
Ok, thanks for your contribution.

PR #779, opened 2022-01-07
Correct german translation of "Power"

rlindner:
The power in watts is normally referred to as "Leistung" in german. "Strom" is used as a synonym of "Stromstärke" for the amperage.


exelban:
Hi. Thanks.

exelban:
I have just noticed that there is a PR with this fix: #770

PR #816, opened 2022-01-27
Update reader.swift

elieserme:
Make compatible with RadeonSensor and SMCRadeonGPU


elieserme:
Please cancel, it's wrong.

PR #864, opened 2022-03-04
feat: Added "Fastest Fan" value in calculated sensors.

fireblue:
Hi, I added a "Fastest Fan" value to calculated sensors.
So we can enable "Fastest Fan" in settings, and see the fast fan rpm in the menubar widget.
Useful for people who likes a simple menubar widget with all maxium values like "Hottest CPU" and "Hottest GPU".


exelban:
Hi. Thanks for the contribution.
Looks nice for me.

PR #955, opened 2022-06-05
feat: add option to show sensors with erroneous values

piernov:
Helpful when troubleshooting sensor issues where SMC reports an incorrect value because sensor is missing or has failed.

### DIFF
--- a/Modules/Sensors/main.swift
+++ b/Modules/Sensors/main.swift
@@ -40,6 +40,11 @@ public class Sensors: Module {
             self.popupView.setup(self.sensorsReader.list)
             self.settingsView.setList(list: self.sensorsReader.list)
         }
+        self.settingsView.erroneousSensorsCallback = { [unowned self] in
+            self.sensorsReader.erroneousSensorsCallback()
+            self.popupView.setup(self.sensorsReader.list)
+            self.settingsView.setList(list: self.sensorsReader.list)
+        }
         
         self.sensorsReader.callbackHandler = { [unowned self] value in
             self.usageCallback(value)
--- a/Modules/Sensors/readers.swift
+++ b/Modules/Sensors/readers.swift
@@ -22,9 +22,35 @@ internal class SensorsReader: Reader<[Sensor_p]> {
         }
     }
     
+    private var erroneousSensors: Bool {
+        get {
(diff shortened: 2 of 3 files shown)


exelban:
Hi. Thanks for your PR.
It's nice to open the issue before doing the PR. Why? Because we can discuss if this feature is needed for the app. 

I understand the reason why you need that feature. But for most users it will be unclear what this option means. And I don't think that this feature is really needed in the app.
If you want to debug your "errors" sensors, I can propose you use the SMC app: `/Applications/Stats.app/Contents/Resources/smc list` - to see a list of all sensors
`/Applications/Stats.app/Contents/Resources/smc fans` - fans details

piernov:
I believe I wouldn't be the only one to use the software this way, this is only to be able to recommend it as a better alternative to HWMonitor for this use-case. Either way, even with the feature not eventually merged, people can easily see that the code already exists with a pull request.

cbusillo:
I think this feature is highly useful in the Mac repair community.  So often we know we have a sensor issue due to high fans or bad performance.  Being able to see which sensor isn't reading correctly quickly helps to zone in on the part or board component that is bad.  With pre-2015 Macs, we were able to use MRI.  Currently, we need to compare the reading between the device under test and a known good unit.

exelban:
Ok, I will check this PR. 
I will leave it for some time. If there will be more people who need this, I will merge that.

LogoiLab:
I would find this feature extremely helpful for newer Macs. The current method of sensor hunting is arduous and time consuming. This would reduce mt turnaround time on repairs significantly.

fiveangle:
I agree that this would be highly useful, but i also agree that it should not be the default, to ensure most users do not get confused. I suggest the PR be modified so that the feature is listed at the very bottom of the "Sensors" prefs page below the "Energy" section and entitled "Debug" with the single entry listed something like "Show data for sensors that provide erroneous data (WARNING: this should not be enabled except for diagnostics by technicians)." Thanks for submitting the PR !

muescha:
i suggest that the erroneous sensors should marked in a different color (red)

exelban:
Hi. There saw a long time when this issue was opened. There is no option to merge this PR now.
But now sensors module have a `Unknown sensors` option. And I will add an option similar to this feature with time.

PR #1006, opened 2022-07-20
feat: add M2 MacBooks to the device list

mooons:
Add M2 MacBook Air and M2 MacBook Pro model identifiers to the device list.

Tested on M2 MacBook Air (was showing `Unknown` before the change):

dash

about

M2 MacBook Pro model identifier can be found on the wiki: https://www.theiphonewiki.com/wiki/List_of_Mac_Laptops_with_Apple_Silicon#MacBook_Pro_.2813-inch.2C_M2.2C_2022.29

It would be great if someone with a M2 MacBook Pro can confirm that as well. :)


exelban:
Hi. Thanks, LGTM.

PR #1014, opened 2022-07-28
Added Swedish translation

soccypowa:
Created a Swedish translation (Disclaimer: I am not a professional translator).

Where suitable I have used Apples own translations, in other cases I ran with just translating the English  to Swedish.

(As I assume is already known, some strings are not localizable, as for example the refresh interval values, which would be a nice addition for the future)


exelban:
Hi, thanks for your contribution.
Can I add you to the readme as a Swedish translation contributor?

soccypowa:
Of course you can

PR #1152, opened 2022-11-06
doc: add a `--cask` argument to brew snippet

ivankovnatsky:


muescha:
the cask argument is obsolete you can skip it

ivankovnatsky:
Ah, I was just using a nix module and could not find the app among the brews.

PR #1214, opened 2022-12-07
Update German localization

Fighter456:


LostInCompilation:
But you left out quite a few... I think I will also make a PR which updates all untranslated entries.

Fighter456:
@LostInCompilation Just point out/translate the missing ones with the suggestion functionality and I will take them over with having you as co-author.

LostInCompilation:
Oh wow never noticed this is possible. Will do it tomorrow, please leave PR open.

LostInCompilation:
Well doesn't work. I can only suggest changes on lines right around your changes. But nothing else farther away. This PR can be closed and merged, I will make a new one after the merge.

PR #1247, opened 2022-12-27
Determine the CPU chip type at runtime.

HsiangHo:
CPU chip type detection should be at **runtime** not compile time. 
If you use the macro, the value of `isARM` has been determined at compile time.

Here is a scenario:
Someone builds the app on an Intel Mac with no Apple Silicon arch and this app can still work on Apple Silicon Mac using Rosetta 2. 
It does run on an ARM Mac. but, the value of `isARM` is FALSE, not TRUE.

We should better determine the CPU chip type at runtime.

### DIFF
--- a/Kit/types.swift
+++ b/Kit/types.swift
@@ -221,11 +221,13 @@ public extension Notification.Name {
 
 public var isARM: Bool {
     get {
-        var value = false
-        #if arch(arm64)
-        value = true
-        #endif
-        return value
+        var size: Int = 0
+        sysctlbyname("machdep.cpu.brand_string", nil, &size, nil, 0)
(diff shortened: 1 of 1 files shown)


exelban:
Hi. Thanks for the feedback.
I have never noticed that. Even when I build an app on intel based and after running on apple silicon.
Don't need to check that via `sysctlbyname`. This information is already available via SystemKit. Give me know if you will make changes to the PR. Otherwise, I will make this change.

HsiangHo:
AFAIK, `sysctlbyname` is currently the accurate way to detect the real CPU arch on macOS. Other APIs will return incorrect results when running in Rosetta 2. 
You can make this change, please.

PR #1419, opened 2023-04-17
Update Greek translations

vaionicle:
# Changes

- Greek translations

### DIFF
--- a/Stats/Supporting Files/el.lproj/Localizable.strings
+++ b/Stats/Supporting Files/el.lproj/Localizable.strings
@@ -2,7 +2,8 @@
 //  Localizable.strings
 //  Stats
 //
-//  Created by Serhiy Mytrovtsiy on 27/08/2020.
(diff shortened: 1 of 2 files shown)


Stats/Supporting Files/el.lproj/Localizable.strings
```diff
@@ -2,7 +2,7 @@
 //  Localizable.strings
 //  Stats
 //
-//  Created by Serhiy Mytrovtsiy on 27/08/2020.
+//  Created by sudoxcess, Updated by vaionicle on 17/04/2023
```
exelban, line 5: do not change the license. You can only add information about contributors.

vaionicle: @exelban

Ok, Thanks for the info, I will update it and push it back

exelban:
Closed due to inactivity.

PR #1508, opened 2023-06-07
updated hungarian translations

majonez:


majonez:
Hi, There is a missing translation key for the SSD lifetime: "Life".

Issue #1264, opened 2023-01-14
Low power mode status always "Disabled"


effleurager:
Low power mode does not seem to be detected, whether only on battery or always. This has previously worked, but I'm not certain if it's a change in macOS or Stats that's broken it.

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/10201206/212494436-800143bc-78f3-4db2-b264-802939b75605.png">

exelban:
Hi. Thanks for the feedback.
I take a look, and there is a subscription for low-power mode [events](https://developer.apple.com/documentation/foundation/nsnotification/name/1617322-nsprocessinfopowerstatedidchange). But looks like it does not work now.

Looks like apple broke something. Because it works previously for sure. And there is no info about deprecation. The only solution it's waiting till apple will fix that.

Ping me after the next macOS update. I will try to find another way for that if apple will not fix that.

Drallas:
> Hi. Thanks for the feedback. I take a look, and there is a subscription for low-power mode [events](https://developer.apple.com/documentation/foundation/nsnotification/name/1617322-nsprocessinfopowerstatedidchange). But looks like it does not work now.
> 
> Looks like apple broke something. Because it works previously for sure. And there is no info about deprecation. The only solution it's waiting till apple will fix that.
> 
> Ping me after the next macOS update. I will try to find another way for that if apple will not fix that.

@exelban Don't know if Apple is even aware of it, since their battery app show the correct information. I send them Feedback (FB11967793) with the Feedback Assistant. Perhaps it helps.

effleurager:
Updated to Ventura 13.2 (22D49), and this works flawlessly now! 🙌🏽

Drallas:
Running the same version (Ventura 13.2 (22D49)) here but still no response. Did deleted the stats .plist in preferences and restarted, but it's stuck on enabled now if when Lower Power Mode is reported on in macOS.

iamyeizi:
> Running the same version (Ventura 13.2 (22D49)) here but still no response. Did deleted the stats .plist in preferences and restarted, but it's stuck on enabled now if when Lower Power Mode is reported on in macOS.

Something similar happens to me. I'm on v13.2.1 (22D68). When I turn on LPM and it doesn't change that icon to orange until I close and open the app, the same for when I turn off LPM and want that orange fill to go away. IMO it could be something about the cache management of a certain function or the trigger to the system info @exelban 

BTW excellent app ❤️

movinimage:
[...]

PR #1611, opened 2023-09-05
battery: Add "Low Power Mode" to popup

foxtrot:
Hi,

I really enjoy using the Stats app but one of the only things it's missing for me is the low power mode status. I've added the code where I thought it makes sense (I'm not a Swift programmer) and it seems to work for me, but happy to make changes if you're willing to merge.

### DIFF
--- a/Modules/Battery/main.swift
+++ b/Modules/Battery/main.swift
@@ -20,6 +20,7 @@ struct Battery_Usage: value_t, Codable {
     var isCharging: Bool = false
     var isBatteryPowered: Bool = false
     var optimizedChargingEngaged: Bool = false
+    var lowPowerMode: Bool = false
     var level: Double = 0
     var cycles: Int = 0
     var health: Int = 0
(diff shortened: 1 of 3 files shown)


foxtrot:
Aha, I did some looking around and didn't realize this was previously implemented!

#1264 seems to blame Apple for a bug here. I am getting pretty consistent results on my M1 Pro machine running 13.5.1, do you think its possible that it's been fixed by them?

The only thing it doesn't do is update in the popup instantly, but I assume that's just due to polling windows.

foxtrot:
Alternatively, I suppose you could shell out and run `pmset -g` to get the value, but I suspect that is something you've considered, and is a bit clunky.

exelban:
I have macbook pro 16 and mac studio for tests. And cannot find a good way to obtain the low power mode state. I do not remember why but there was some reason to not use a pmset. Will take a look later at what was the reason.

exelban:
Also will recheck if Apple have fixed the `isLowPowerModeEnabled`

Issue #1634, opened 2023-10-01
No way to change network base from bits to bytes if speed widget is disabled.


justinharrell:
If the network speed widget is disabled then there is no way to change the base from bits to bytes since the base setting is on the widget settings.

I use the network chart rather than network speed widget but when clicking on it the popup shows throughput in either bits/s or bytes/s, the default is bytes. I wanted to change this to bits and could not find the setting so I thought it was not a feature, so I searched existing issues and found it is there but you have to enable the networks speed widget, change it, then it can be disabled again and it changes the popup base as well.

Seems to be the network base setting should be in the module settings not the widget settings or the widget and popup should have independent base settings (which doesn't make sense to me). Looks like this may have caused confusion for others in the past based on the requests to have the feature even after it was implemented.

PR #1635, opened 2023-10-01
Moved network base setting to module

justinharrell:
From issue #1634

The network base could not be modified if not using the network speed widget even though the popup also uses that setting.

This moves the setting from the widget to the module.

I put the setting at the top above others as that made the most sense to me, seems like a key module setting.

Also fixed what looked like a typo on the icon setting default value.

### DIFF
--- a/Kit/Widgets/Speed.swift
+++ b/Kit/Widgets/Speed.swift
@@ -15,7 +15,6 @@ public class SpeedWidget: WidgetWrapper {
     private var icon: String = "dots"
     private var state: Bool = false
     private var valueState: Bool = true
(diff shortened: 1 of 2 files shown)


exelban:
thx. 
PS: you forgot to add this option to the Drive module. Not only Network module using this widget.

justinharrell:
Ah yes, I have not used the disk speed widget and missed that separate base setting, probably wouldn't have caught it either way since I am used to measuring disks in byte/second not bit/sec which is much more common in networking. 

Thanks for accepting and all the hard work on Stats!

exelban:
np. I have added that already.
thx.

Issue #1623, opened 2023-09-22
[Feature Request] Right-click module for more details in Combined Mode


kevllarco:
Thanks for developing Stats! I love it. I like using it in Combined Mode as it's easier to manage and move around. The only problem is you lose a lot of the details in the dropdown menus. Is there a way you could allow users to ctrl-click or right-click to get access to the full module detail?

For example, while in Combined Mode, the user would have to ctrl-click or option-click to get the full details of the CPU module and see all the processes running. I think a direct keyboard shortcut for each module would resolve this issue also but I didn't see any option for that. Thanks so much!

justinharrell:
Definitely second this seems duplicate of #1604, #1497, #1353, #1449 and probably more. 

Seems like it should be relatively simple to display the separated popup next to the combined popup. iStat does this on hover on the combined module, so if you mouse over cpu the separated cpu pane pops next to it. Hover can be a little aggressive I would be fine with just a single click popping the separated details up and of course being able to drag to keep it open.

The other issue with combined is not being able to hide widgets while still seeing them in the popup, that is on the menu bar I just want to see battery cpu and network, but if I click on it then see all the other modules.

mailinglists35:
hi @exelban 
why is this not planned? is it technically impossible?

PR #1639, opened 2023-10-02
Full popups can now be accessed in combined mode 

justinharrell:
Clicking on the modules portal in the combined modules popup will show the full popup for that module. This allow seeing full details even in combined mode.

Dragging to pin still works and has been improved such that clicking on widget or portal will bring pinned window to front rather than resetting.

Based on issue #1623. I choose to show on single click which feels most natural to me rather than right click or other modifier. Very quick and intuiative IMO allowing drill down into details with minimal changes to codebase.

### DIFF
--- a/Kit/module/module.swift
+++ b/Kit/module/module.swift
@@ -290,7 +290,11 @@ open class Module: Module_p {
             popup.openedBy = widget
         }
         
-        if popup.occlusionState.rawValue == 8192 || reopen {
+        if(popup.locked){
+            popup.orderFront(self)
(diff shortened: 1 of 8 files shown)


paulora2405:
Are there any updates on maintainers reviewing this PR?

exelban:
Hi. Sorry for no responding for a long time.
Nope, I will not going to merge this PR. Because I don't like this solution. This combined module was just a quick solution for users who are using the merged widgets option. I don't like how it looks now, and I have in plan to redesign it from scratch. But cannot find a time for that(
This PR it's just a quick fix that I don't like and does not fit how I see this feature.

PR #1701, opened 2023-11-25
Update readers.swift

Catalin9090:
Add temperature average for all sensors and for CPU+GPU


exelban:
hi. No, I will not add custom sensor calculations. If you need it you are free to use your version of Stats.

It must be a mechanism that allows the user to select any sensors and show them as sum/average/etc.

Catalin9090:
Ok. I don't need it. I wanted to try and "help" the community. Thank you for the fast response !

PR #1746, opened 2023-12-27
Initial implementation of custom scale for network charts. WIP. Fixes #844.

### LINKED ISSUE #844: [Enhancement] Network graph widget enhancements proposal
Hi.
I like to use the network graph, but I am disturbed by the "auto-scale" of the current implementation. Would it be possible to have a setting to use a fixed value (maybe based on the interface speed, or user selectable)? 
Another change would be to have the possibility to use a logarithmic scale for the graph (as in MenuMeters).

victor-marino:
Please note, this is my first time ever developing for MacOS, *or* using Swift, *or* even using Xcode for development 😅

As a result, I apologize in advance for any errors or terrible practices I may have used here!

What's implemented:
* New toggle to enable/disable the Custom Scale option
* New text field to enter the maximum desired bandwidth for the chart (in Mbps)
* Settings added under the Network settings panel for both the widget and the pop-up charts
* Both charts will pick up the new value as soon as the text field is modified (if the toggle is enabled)

Things I couldn't figure out how to do by myself:
* Hide or disable the input text field when the toggle is disabled. Looked up some state management tutorials but wasn't able to make it work here.
* Sanitize the input of the text field to allow only numerical digits.
* Allow entering separate values for the download and upload speed. Couldn't figure out how to identify which text field was notifying the changes, so I went with a single text field for now.

Hope this is at least helpful as a starting point ;-)

### DIFF
--- a/Kit/Widgets/NetworkChart.swift
+++ b/Kit/Widgets/NetworkChart.swift
@@ -10,8 +10,9 @@
 //
 
 import Cocoa
+import SwiftUI
 
-public class NetworkChart: WidgetWrapper {
+public class NetworkChart: WidgetWrapper, NSTextFieldDelegate {
     private var boxState: Bool = false
     private var frameState: Bool = false
     private var labelState: Bool = false
@@ -20,6 +21,8 @@ public class NetworkChart: WidgetWrapper {
     private var uploadColor: Color = .secondRed
     private var scaleState: Scale = .linear
     private var commonScaleState: Bool = true
+    private var customScaleState: Bool = false
+    private var customMaxBandwidth: Int = 0
     private var reverseOrderState: Bool = false
     
     private var chart: NetworkChartView = NetworkChartView(
@@ -85,6 +88,8 @@ public class NetworkChart: WidgetWrapper {
             self.uploadColor = Color.fromString(Store.shared.string(key: "\(self.title)_\(self.type.rawValue)_uploadColor", defaultValue: self.uploadColor.key))
             self.scaleState = Scale.fromString(Store.shared.string(key: "\(self.title)_\(self.type.rawValue)_scale", defaultValue: self.scaleState.key))
             self.commonScaleState = Store.shared.bool(key: "\(self.title)_\(self.type.rawValue)_commonScale", defaultValue: self.commonScaleState)
+            self.customScaleState = Store.shared.bool(key: "\(self.title)_\(self.type.rawValue)_customScale", defaultValue: self.customScaleState)
+            self.customMaxBandwidth = Store.shared.int(key: "\(self.title)_\(self.type.rawValue)_customMaxBandwidth", defaultValue: self.customMaxBandwidth)
             self.reverseOrderState = Store.shared.bool(key: "\(self.title)_\(self.type.rawValue)_reverseOrder", defaultValue: self.reverseOrderState)
             
             if let downloadColor =  self.downloadColor.additional as? NSColor,
@@ -92,6 +97,7 @@ public class NetworkChart: WidgetWrapper {
                 self.chart.setColors(in: downloadColor, out: uploadColor)
             }
             self.chart.setScale(self.scaleState, self.commonScaleState)
+            self.chart.setCustomScale(self.customScaleState, self.customMaxBandwidth)
             self.chart.reinit(self.historyCount)
             self.chart.setReverseOrder(self.reverseOrderState)
         }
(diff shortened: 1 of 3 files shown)


victor-marino:
@exelban, did you get a chance to look into this? Any feedback?

Regarding the To-Do's:
If you don't have the time to work on the pending UI stuff, I was thinking maybe we could simplify it by removing the toggle altogether. Instead, the feature could be turned on or off just based on whether the user enters a value in the text field or leaves it empty (I've seen you've done this with other settings in the app).

Let me know what you think 😉

exelban:
I don't have a time. Will take a look next week. But I'm not sure I understand what do you want to archive with feature.

victor-marino:
Great, thanks a lot.

> But I'm not sure I understand what do you want to archive with feature.

I'll try to explain.

Right now, Stats will always modify the scale of the network charts dynamically. Meaning, if your maximum DL speed for the visible period is, say, 10 Kbps, the maximum value of the chart will be 10 Kbps. If there's a traffic spike to something like 100 Mbps (during a download), the chart will scale to that value, and so the 10 Kbps traffic will become invisible (as it should be... it was probably just background traffic). But once that 100 Mbps spike leaves the chart, it will suddenly scale back to whatever the new "maximum" is, and those little 10 Kbps spikes will again take over the whole chart.

This behaviour is fine for certain use cases... but is *really* bad for others.

First of all, it's frustratingly inconsistent. The chart being maxed out could mean I just downloaded a full GB of data, and a few minutes later, that same chart being maxed out could just be showing some background traffic. Without looking at the scale, the plot tells me nothing because it isn't consistent. Having to double check the scale every second makes for a sub-optimal UX in my opinion, often making things confusing.

Second, because the widget chart doesn't have a scale, that particular chart tells you almost nothing all the time. You can only rely on the actual speed numbers, but not on the chart.

Third, as a consequence of the previous point, you also lose the ability to use the widget chart as a handy "warning" for abnormal traffic. Right now, I can't assume that a maxed out chart means I'm actually downloading anything, because the chart is always maxed out!

And finally, you also lose the ability to use the chart as a measure of how _congested_ your internet connection is. Any traffic will put it at 100%, so it looks as if your internet connection is always fully utilized, even when there's only background traffic.

For these reasons, many traffic monitor apps allow you to set a fixed scale, which you will often want to set as the maximum speed of your home broadband connection (e.g.: 1 Gbps). Here's a screenshot from DU Meter, probably the most popular traffic monitor app for Windows:
imagen

If you think of your internet connection as a "water pipe", setting the scale to match your max broadband speed will turn the chart (and especially the widget chart) into a graphical representation of how full that pipe has been for the past few seconds. This way, the chart stays pretty much empty unless I'm actually doing some significant traffic, which is exactly what I want. In other words, I want the chart to act as a **percentage** representation of my internet connection usage. So if I'm downloading at 1 Gbps on my 1 Gbps fiber connection, I want the chart to be 100% full. And if I'm downloading at 500 Mbps, I want the chart to be at 50%.

I'm a network engineer myself, and I can't tell you how useful it is to just glance at the chart and know how congested my internet connection is without having to even look at the numbers or click anything. For instance, if I'm browsing the web and things are slow, a quick glance at the chart may already show my pipe has been full for a while, so no need to even open the pop-up. For me it's really the number 1 feature I go looking for when using any traffic monitoring app.

I understand this may not be the desired behaviour for everybody, which is why it should be an option. But I hope you can at least understand the value of having including it 😉

victor-marino:
Hi.

> Sorry, but have no time this week.
No worries at all. I really appreciate you looking into it whenever you can, and I'm just happy to contribute to the project and improve the app.

> There are 2 options now: you can finish the PR by doing all TODO or transform this PR to the issue and I will implement this feature when will have time.

We will have to go with option 2 unfortunately. I'd love to help more, but as said I'm a complete noob in MacOS development and I simply don't know how to complete the missing parts. I think it _should_ be relatively simple though (mostly some UI tweaks), but I think it's best if you do it yourself, whenever you get some time.

What do you mean by "transform" this into the issue? Just reopen the issue so you can track it properly? Or is there anything else I should do?

> Also what about the situation when io goes over the limit? How to represent that if there is be locked max value?

The chart will simply be maxed out (saturated, completely filled). This is what happens already with my current implementation in the PR.

I don't think that's a problem, as it's how all apps handle it, or at least all the apps I've seen. It's probably what users will expect when enabling this feature.

exelban:
it will be enough to open the issue with a proposal. And copy past your messages there. Also you can notice that PR.

What do you mean the chart will be maxed out? If there will be a fixed max value how could it go out from that range?

victor-marino:
> it will be enough to open the issue with a proposal. And copy past your messages there. Also you can notice that PR.

Ok, will do 👍🏻

> What do you mean the chart will be maxed out? If there will be a fixed max value how could it go out from that range?

Sorry, I didn't explain myself properly. By "maxed out" I mean the chart will simply be full, at 100%.

If the scale is set to 1 Gbps but your throughput is 1.1 Gbps, then the graphical representation will be the same as if the throughput was 1 Gbps. The numeric values will allow you to see the real value of course.

If you test my PR you'll see the chart simply goes to 100% in this case.

The nice thing is, if the user expects this to happen frequently (e.g.. some ISPs "overprovision" fiber speeds), you can always set the max value to be something like 10% higher than your "official speed" (1.2 Gbps in this case).

victor-marino:
New issue created: https://github.com/exelban/stats/issues/1787

PR #1756, opened 2024-01-02
lang: Update Spanish localization

victor-marino:
I noticed the Spanish localization was quite outdated, and some of the existing translations were also not very accurate. I'm a native Spanish speaker, so I decided to help ;-)

All strings in v2.9.18 are now translated in this PR. I made sure they look good in the UI too (shortened some of them to make them fit).


Stats/Supporting Files/es.lproj/Localizable.strings
```diff
@@ -7,7 +7,7 @@
 //  Running on macOS 10.15.
 //
 //  Copyright © 2020 Serhiy Mytrovtsiy. All rights reserved.
-//
+//f
```
Fighter456, line 10: This change seems to be unintentional.
```suggestion
//
```

Fighter456 (changes requested):
The mentioned line seems to be changed in an unintentional way.

Stats/Supporting Files/es.lproj/Localizable.strings
victor-marino: Apologies for that, probably a failed `CMD+F` 😅

Fighter456 (approved):

PR #1852, opened 2024-03-14
lang: updated Simplified Chinese translation

Jerry23011:


PatShen (commented):
There some other strings which haven't been translated, could you please also update them?

* No notifications available in this module
* Left fan
* Right fan
* Fastest fan

Jerry23011:
thank you for reminding :)

PatShen (approved):

Issue #606, opened 2021-09-06
nettop high CPU usage


dima-xo:
**Describe the bug**
Nettop uses a lot of cpu.

Sample of stats:
[Sample of Stats.txt](https://github.com/exelban/stats/files/7113633/Sample.of.Stats.txt)

Parent process — stats.
![Screen Shot 2021-09-06 at 08 53 57](https://user-images.githubusercontent.com/28823624/132167530-7f79730e-468f-4aea-b623-97a9aba9f4ec.png)
![Screen Shot 2021-09-06 at 08 54 19](https://user-images.githubusercontent.com/28823624/132167539-b7176391-3e16-40f3-89bc-109c45359c1e.png)

**Details:**
 - Device: MacBook Pro m1
 - MacOS: 11.5.2
 - Application version: 2.6.7

dima-xo:
2.6.8 same results

exelban:
Hi. Thanks for the reporting.
`nettop` must running only when you have opened the `Network` details window. Does the nettop run always (even when you do not have any Stats window opened)?

exelban:
Could you please check what `Reader type` you have selected in the `Network` module settings?

dima-xo:
> Hi. Thanks for the reporting.
> `nettop` must running only when you have opened the `Network` details window. Does the nettop run always (even when you do not have any Stats window opened)?

Yes, unfortunately. It is always launched, sometimes 2-3 instances at a time.
When minimized, the statistics do not work. Always shows up: 0, down: 0.

> Could you please check what Reader type you have selected in the Network module settings?
Process based. If I choose "interface based" with or without autodetection, the same result — nettop doesn't unload from memory.

(I'm using a VPN through wireguard if that helps, but before (a few versions ago, I don't remember when exactly) everything was stable)

exelban:
Hmm, so even when you have selected `interface based` the `nettop` still runnning?
(VPN do not have an impact, it could only provide wrong stats)

dima-xo:
Yes, I can describe the case:

Offload the nettop from memory (via force quit).
I enable Network in stats with interface based / process based.
Nettop is not running.
I click on network on top bar for details.
Nettop is up and running.
I hide the network's details.
nettop stays in memory.

It's worth noting that when Interface based is enabled, the stats on the top panel are updating (when Process based is enabled, zeros are displayed) 

nettop is triggered by a detailed statistics of the network. But it doesn't offloaded from memory.

dima-xo:
[...]

Issue #536, opened 2021-07-11
A lot of CPU load by nettop


linkinek:
**Describe the bug**
I have a lot of CPU load by nettop process. It it happening after update to latest version and happens in randomly time.
CPU normalise only after close Stats and kill nettop process. And if I kill nettop process without close application, process starts again with a lot of CPU load.

**Details:**
 - Device: Macbook Pro 2018
 - MacOS 11.4 (20F71)
 - Application version 2.5.22

exelban:
Hi. Thanks for reporting.
Nettop is using to fetch top processes. And it must be running only when you open the popup view. Could you please run the Stats and do not click on the items in the menu bar at all. After some period of time check in the Activity Monitor if the problem appears.

linkinek:
Hi. 
I updated to 2.5.60 and problem exists yet. Nettop starts immediately after starting the application. But now there are 2 of them.
Can I attach some information for analyse?

exelban:
Hi. What type of network reader do you use?

linkinek:
What do you mean?

exelban:
There is an option `Reader type` in the network module. What option you have selected?

linkinek:
Process based

exelban:
Oh, now I see. Unfortunately, it cannot be fixed. 
If you have selected `Process based`, the app calculates network traffic based on `nettop` data. I could only recommend switching to the `Interface based`.

`nettop` is an internal macOS tool. And it's only one way to fetch network traffic without root permissions.

linkinek:
Oh... Is there a difference between Process based and Interface based?

exelban:
Yes. `Process based` using `nettop` and calculating network bandwidth based on the sum of all processes.
`Interface based` listening to the bandwidth on the specific interface which is active.

linkinek:
Ok. Thanks. I think can close the issue.

Issue #216, opened 2020-12-07
Suspicion of system resources consumption forcing to reboot


bouscram:
**Describe the bug**

**_At the moment, I'm not sure that Stats is the cause, but things have been OK since I stopped it_**

I have seen since a few days (one week or a bit more), a strange phenomenon happening. At some point in time, it seems that OS resources are exhausted, which has several effects:

- Safari says it is missing essential parts and needs to be reinstalled
- External disks are not recognised anymore
- System is very slow and Stats' CPU widget shows constant activity (sometimes up to 100%)
- Apps are slow to start or can't even be launched

Memory consumption looks normal.

When this happens, the only thing possible is to reboot the machine.

I have stopped Stats since Saturday and things seem to be normal again.

**Details:**
- Device: end 2012 iMac
- MacOS: 10.15.7
- Application version: at least starting with 2.4.5
- Intel PowerGadget installed

exelban:
Hi. Thanks for the feedback.
It's really strange situation.

These points could not be related to the app: 
```
- Safari says it is missing essential parts and needs to be reinstalled
- External disks are not recognised anymore
```

Could you please check the CPU usage of Stats?
Will be great if you can post the top CPU processes from the popup view.

Maybe there's some leak, and it can be the reason for 100% CPU usage.

bouscram:
I must say I am very puzzled. CPU usage was generally higher than normal (about 25/35% when doing nothing) but I didn't see any unusual process in the list. Stats wasn't using more CPU that usually, i.e. less than 1%.

I guess that the Safari and disk problems were only side effects, like if a resource they need was not available. Memory seemed OK but I didn't check the size of the virtual memory used (I should have).

For the moment I have stopped Stats (at least for a few days, to be fairly sure that the issue was related). If I have some time, I'll do more testing and will keep you posted.

exelban:
There is an option that Stats consume a lot of CPU. If some leak happened. But I hope I catch all of them.

I'm very astonished by this behavior.
```
- Safari says it is missing essential parts and needs to be reinstalled
- External disks are not recognised anymore
```

This can not be related to CPU or Memory consumption. It looks like some problem with the system.

Because, for example, these two could be related to each other:
```
- System is very slow and Stats' CPU widget shows constant activity (sometimes up to 100%)
- Apps are slow to start or can't even be launched
```

bouscram:
Hi. After one week without Stats (where I didn't see any issue), I restarted this morning.

[...]

PR #1854, opened 2024-03-15
fix nettop may occupy CPU abnormally

caobug:
This issue may occur if we don't set `NSUnbufferedIO = YES` and `standardInput`. When `nettop -L = 0`, it's easy to reproduce.

https://github.com/exelban/stats/issues/606
https://github.com/exelban/stats/issues/536
https://github.com/exelban/stats/issues/216


exelban:
LGTM. Thanks for your contribution.

PR #1879, opened 2024-04-05
[fix] fix Sensors align in combined view and added localizedString to PortalHeader

ShigemoriHakura:


exelban:
LGTM. Thanks for your feedback.

PR #1936, opened 2024-05-08
Change "Model id" string to "Model identifier"

infinitepower18:
Updated the "Model id" localized string to "Model identifier". Apple officially uses the term model identifier so it makes sense to have the app reflect this.


exelban:
thx.

PR #1945, opened 2024-05-15
Improve zh hant

jack24254029:
1. Add Import / Export settings of localizable
2. Update localizable of zh-Hant


exelban:
LGTM.
Thanks for your contribution.

Issue #1908, opened 2024-04-22
Combined modules not displaying


theangrydev:
![image](https://github.com/exelban/stats/assets/6877541/9d8bea18-6564-47e3-8916-1b4b4bafc3b2)
[Stats.plist.zip](https://github.com/exelban/stats/files/15055919/Stats.plist.zip)

eric-saintetienne:
You're lucky to get all of them because even when not combining them I get only a couple of them displayed, somtimes the batter, sometimes the network, sometimes the cpu or the bluetooth, but never all of them like you.

exelban:
Hi. Hmm, strange. I have imported your settings and everything works fine. Need to debug that. If you can reproduce that problem at any time I can make a debug build that will tell more what is going on.

theangrydev:
@exelban sure happy to retry with a debug build, this reproduces consistently for me

exelban:
hi. Thanks for your help.
There is a debug build:
[Stats.dmg.zip](https://github.com/exelban/stats/files/15213035/Stats.dmg.zip)


Please run this build. It will collect some logs to the `Documents/log.txt`. Please share these logs afterward.

PR #1947, opened 2024-05-16
fix: change log file directory to "Application Support"

minimalic:
Stats currently generates a `log.txt` file inside local "Documents" directory. It's not a place to store application related data.

I suggest we use the "Application Support/Stats" directory inside local "Library" for the `log.txt`.


exelban:
hi, thanks for the PR. But there is no reason for that. 
There must be no logs and it appears just because I have added that to debug #1908. And forgot to remove it(

I have already updated the release with disabled logs.

PR #1985, opened 2024-06-09
Add English (U.K.) language variant

Coopydood:
Hey, love the project, and wanted to do my part. I'm from the UK, and we have some slight language differences from standard English (US). 

I've added a Localizable file for English (U.K.) / `english_gb`. It covers the few strings that differ from the US variant, namely:

Color > Colour
Utilization > Utilisation
Gray > Grey
Center > Centre

image

Thanks for your work, it's appreciated.

Take care

### DIFF
--- a/Stats/Supporting Files/en_gb.lproj/Localizable.strings
+++ b/Stats/Supporting Files/en_gb.lproj/Localizable.strings
@@ -2,7 +2,7 @@
 //  Localizable.strings
 //  Stats
 //
-//  Created by Serhiy Mytrovtsiy on 27/08/2020.
+//  Translated with <3 by Coopydood
 //  Using Swift 5.0.
 //  Running on macOS 10.15.
 //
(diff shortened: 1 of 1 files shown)


Coopydood:
What needs to be changed, @exelban? Not sure why the check is failing :/

Stats/Supporting Files/en_gb.lproj/Localizable.strings
```diff
@@ -0,0 +1,451 @@
+//
+//  Localizable.strings
+//  Stats
+//
+//  Translated by Coopydood
+//
```
exelban, line 6: you removed these lines from the header:
//  Using Swift 5.0.
//  Running on macOS 10.15.

and now the line number is not correct

exelban:
Please feel free to leave the author's information on line 5.

Coopydood:
Ready to go when you are :)

PR #2004, opened 2024-06-22
Tweak GHA configs a bit

Goooler:
- Bump actions to the latest, the old ones based on Node 16 are deprecated.
- Migrate build devices to macos-latest that now based on Apple Silicon Chips, will significantly speed up builds.


exelban:
hi. Thx for PR. But I don't have any reason to do that. Actions work normally, and I don't want to change since it works fine.

PR #2090, opened 2024-08-25
fix: make log scale graph have better dynamic range

### LINKED ISSUE #2088: Suggestion: rescale logarithmic network graph to show 1 kB/s closer to y-axis
I love that you have logarithmic scaling for the network graph. That's a great touch!

One suggestion: right now, there's unfortunately relatively little dynamic range because the y-value for 1kB/s is already relative large:

image

It would be great if you changed the graph log y-scale so that 1kB/s was closer to the 0-value of the y-axis. This would easily double the dynamic scale.

You might feel like this is a hack - but it isn't! You always have to arbitrarily choose where to put the minimum value on a log scale that doesn't go all the way to 0. Right now, the y-axis is maybe at around 1B/s which is never actually measured. The minimum that can be resolved is 1kB/s in my experience, so that's what should be close to 0 (not exactly 0 because we want to be able to tell 1kB/s apart from 0kB/s).

Does my suggestion make sense? I'm happy to explain further if it's not clear.

corneliusroemer:
Resolves #2088

Previously, dynamic range of log scale was bad because 0.01 Bytes/second (10e-2) was mapped to 0

Commonly found values for network speed range from 1kB/second (10e3) to 100GB/second (10e8)

As a result, half of the dynamic range was wasted as no values would ever appear between 10e-2 and 10e3.

New:

image

Before:

Stats 2024-08-25 19 21 59

### DIFF
--- a/Kit/plugins/Charts.swift
+++ b/Kit/plugins/Charts.swift
@@ -45,8 +45,6 @@ internal func scaleValue(scale: Scale = .linear, value: Double, maxValue: Double
             localMaxValue = cbrt(maxValue)
         }
     case .logarithmic:
-        // Smallest value passed is 1024
-        // Setting y=0 to 256 gives a nice curve
         let zeroVal = 256.0
         if value > 0 {
             value = log(value/zeroVal)
--- a/Kit/plugins/Charts.swift
+++ b/Kit/plugins/Charts.swift
@@ -45,11 +45,14 @@ internal func scaleValue(scale: Scale = .linear, value: Double, maxValue: Double
             localMaxValue = cbrt(maxValue)
         }
     case .logarithmic:
+        // Smallest value passed is 1024
+        // Setting y=0 to 256 gives a nice curve
+        let zeroVal = 256.0
         if value > 0 {
-            value = log(value*100)
+            value = log(value/zeroVal)
         }
         if localMaxValue > 0 {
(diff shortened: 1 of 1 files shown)


Kit/plugins/Charts.swift
```diff
@@ -45,11 +45,14 @@ internal func scaleValue(scale: Scale = .linear, value: Double, maxValue: Double
             localMaxValue = cbrt(maxValue)
         }
     case .logarithmic:
+        // Smallest value passed is 1024
+        // Setting y=0 to 256 gives a nice curve
```
exelban, line 49: please remove that comments

corneliusroemer: Sure

corneliusroemer, line 49:
```suggestion
```

Kit/plugins/Charts.swift
```diff
@@ -45,11 +45,14 @@ internal func scaleValue(scale: Scale = .linear, value: Double, maxValue: Double
             localMaxValue = cbrt(maxValue)
         }
     case .logarithmic:
+        // Smallest value passed is 1024
```
corneliusroemer, line 48:
```suggestion
```

Kit/plugins/Charts.swift
corneliusroemer: Done

Issue #2008, opened 2024-06-25
Battery widget doesn't show percentage inside the icon when charging


Gleek:
**Describe the bug**
For the horizontal battery widget, "Percentage inside the icon" option is available in Additional information.
<img width="542" alt="image" src="https://github.com/exelban/stats/assets/3756804/9d4f5246-396b-4fbf-91f5-cffdcaa6e918">

This works correctly when the device is not connect to power
![image](https://github.com/exelban/stats/assets/3756804/61c9bfac-9ddf-4014-b766-a4bd9a484cd5)

But hides the text and replaces it with a charging icon when the device is connected to power.
![image](https://github.com/exelban/stats/assets/3756804/778635ab-f13b-448f-b485-9a378123131e)

An icon like the following could be used here instead, that would satisfy having a charging indicator along with percentage inside the icon 
![image](https://github.com/exelban/stats/assets/3756804/a9d3d01f-323e-446e-a643-465ffa15c5b8)

**Details:**
 - Device: Macbook Pro 2023
 - macOS: 14.5
 - Application version: 2.9.21

koushik-chowdhury:
I noticed that my pull request (#2008) was closed without any feedback. I would really appreciate it if you could provide some insight into the reason for the closure. I’m eager to contribute to the project and would love to understand how I can improve my submission to better meet the project's standards or expectations.

cristiean-travelex:
I suggest the charging icon to appear inside the battery icon, before or after the digit characters. This will ensure consistent icon width.

gordonbazeley:
Somewhat related to this - I'd rather use the vertical bar module for the battery using combined module icon in the menu bar. Would it be possible to include some visual representation showing whether the battery is charging within the popup control when pressing on the combined module icon please?

PR #2118, opened 2024-09-12
Issue #2008 Display Charging Percentage Instead of Power Icon

koushik-chowdhury:
This pull request resolves Issue #2008, where the power icon was displayed when the device was connected to power. The modification ensures that the charging percentage is shown instead of the power icon when the device is plugged in.

### DIFF
--- a/Modules/Battery/main.swift
+++ b/Modules/Battery/main.swift
@@ -119,9 +119,9 @@ public class Battery: Module {
                 widget.setValue([[ColorValue(value.level)]])
                 widget.setColorZones((0.15, 0.3))
             case let widget as BatteryWidget:
-                widget.setValue(
+                 widget.setValue(
                     percentage: value.level,
-                    ACStatus: !value.isBatteryPowered,
+                    ACStatus: false,  // Always false to avoid showing the power icon
                     isCharging: value.isCharging,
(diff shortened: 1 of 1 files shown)


exelban:
@koushik-chowdhury you literally just remove the function from the app. I don't event know how to commend that.

koushik-chowdhury:
@exelban Thank you for your response, and I apologize if my changes caused confusion or removed an important function unintentionally. That certainly wasn’t my intention. I appreciate your feedback, and I’d like to better understand what went wrong so that I can correct the issue and resubmit a more suitable pull request.

If you have any guidance on how I can reintroduce the functionality and align with the project’s goals, it would be greatly appreciated. I’m eager to improve my contribution and make it useful for the project.

exelban:
In the case of #2008 there is a need to rewrite the battery widget to move the lighting icon to the side as a feature flag.
I'm still not convicted if it's needed. Since I try to keep the battery widget similar to the native one.

koushik-chowdhury:
@exelban Thank you for the clarification. I understand your preference to keep the battery widget similar to the native one, and I agree that maintaining consistency is important. I believe your direction is the right approach for the project, and I’m happy to follow it.

Thanks again for your feedback!

Issue #479, opened 2021-05-14
REQ: bolder font for text based items


gingerbeardman:
**Describe the feature request**

Bolder font option for items such as network I/O

**Details:**
 - Device: Mac mini (Late 2018)
 - MacOS 11.3.1
 - Application version 2.5.14

**Alternative**
iStat Menus, where I'm coming from, offers this.

**Note**
I am happy to file a PR for this if it would be accepted.

exelban:
Hi. Thanks for the proposal.
I'm not sure if it could be a boulder text. I'm trying to use the max font size and weight which will be drawn normally on different screens (dpi and resolutions). But if you have an idea of how it can be realized, you can make a PR and we will see if it's acceptable)

gingerbeardman:
I'll have a look/think about this! Thanks

exelban:
Hi. There are some changes?

gingerbeardman:
Sorry no, I don't think I will have time.

exelban:
Ok, I'm going to close this issue. It could be reopened if someone will have an idea of how it could be done.

gingerbeardman:
OK, somewhat disappointing.

gingerbeardman:
Here's a comparison of iStats menus vs Stats

<img width="140" alt="image" src="https://github.com/user-attachments/assets/2a8146fc-7614-4576-ad02-9a6a206264aa">

PR #2169, opened 2024-10-07
Bolder text for speed module menu bar item

gingerbeardman:
- bolder text for speed module menu bar item

I guess this asks questions
- whether the same changes should be applied to all modules
-  if it should be an option

Given that it changes existing visual look and feel, perhaps a "bolder text" option could work? then each use of NSFont.Weight would need a ternary conditional and pick between two different weights. Up to you, to be discussed.

This small changes is all I needed for my local copy as I only use the speed module.

## Before
Screen shot 2024-10-07 at 15 53 55

## After
Screen shot 2024-10-07 at 15 53 31

The 0.0 and 0.0 is iStats Menu "Font = Bold" for comparison sake.

And here's standard Stats versus iStat Menus "Font = Regular", Stats is still lighter.

Screen shot 2024-10-07 at 15 56 32

In addition, the exact size and alignment of the text could be improved to minimise aliasing and increase legibility.

## Fixes
- #479

### DIFF
--- a/Kit/Widgets/Speed.swift
+++ b/Kit/Widgets/Speed.swift
@@ -223,7 +223,7 @@ public class SpeedWidget: WidgetWrapper {
         let size: CGFloat = 10
         
         let downloadStringAttributes = [
-            NSAttributedString.Key.font: NSFont.systemFont(ofSize: 11, weight: .regular),
+            NSAttributedString.Key.font: NSFont.systemFont(ofSize: 11, weight: .medium),
             NSAttributedString.Key.foregroundColor: color,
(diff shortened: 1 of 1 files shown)


exelban:
Hi. I have checked this. And I totally do not like that. I understand that someone needs to have bolder text for some reason in the menu bar. But for example I don't. I like to have menu bar as much clean as I can. So it will be not changed for sure.
But I will agree if there will be a toggle option to specify if the text must be thin or bold.

gingerbeardman:
Understood!

PR #2189, opened 2024-10-23
adjust Text parser

luckman212:
This is a proposed change to the Text parser.

The current implementation does not handle strings where the placeholder is directly adjacent to another char, e.g. `(`

E.g. this:
```
$type $interface.displayName ($interface.BSDName)
```

This changes it to use regex that does allow this.


exelban:
LGTM. Thanks for your contribution.

PR #2270, opened 2024-12-15
Added Language: English Australian

Will-Georges:
Added English Australian Localisation.

Feel free to let me know if I need to make any changes.


exelban:
Hi. Sorry for the question, but what the difference between English and English Australian?
I take a quick look and find no difference in the translation.

Will-Georges:
There aren't many differences:

- Color is colour
- Utilization is utilisation
- Gray is grey

If it not worth the change then don't bother, but they are the main changes.

Will-Georges:
I just thought that since you also have en-GB which is similar, you may as well have en-AU as well.

exelban:
It's not a problem to add, but will be problem to support. If you ok with monitor and support it ok, otherwise I don't think it's worth to add that for 3 words (which also very similar)

Will-Georges:
When you say support it, do you mean when new words and things are added, or what exactly?

If it is just for when new words come then I can do that. I use the app all the time so I'm happy to help.

Will-Georges:
And also I am a bit of a code noob so I don't know how to resolve the conflict. Like I see what the problem is, but I don't know how to fix it through GitHub, if you could either do it or teach me please?

PR #2274, opened 2024-12-16
Add M1 / M2 Mac Mini Icon

ben-wheeler:
Always noticed that the M1 / M2 Mac Mini shows the older Intel Mac mini icon on the dashboard.

I've simply added the M1 / M2 Mac Mini icon, as taken from system files and compressed down to match the size of other icons.

Before
image
After
image

The Intel 2018 model will still show its respective icon. I was tempted to make the silver icon the default fallback, (as it's very similar to the 2014 and 2012 model icon). However, I'm unsure if the app still functions on older versions of macOS, so this change might not even roll out to any of those machines. Happy to swap the fallback over if desired.


exelban:
LGTM, thanks for your contribution

Issue #2162, opened 2024-10-03
[Feature request] Aggregate RAM usage by process


yvesrichardhth:
In the ram widget, would it be possible to have an option to aggregate the RAM usage by process ? In macOS "Safari web content" shows many times, and it would be great to have the sum of all the same processes names in order to know the global amount of RAM usage.
<img width="293" alt="image" src="https://github.com/user-attachments/assets/e46fc27d-7dd6-45d4-8430-ad422cc06243">

ikovacev:
+1, this is the only way of seeing the true Safari memory impact which is the top memory user for me.

Issue #2099, opened 2024-08-31
RAM usage: total usage per main process


TheBlinded:
**Describe the bug**
When looking at RAM usage, especially for browsers, I expect to see the overall usage of a program and not split into its many sub-processes. See screenshot.
![Schermata 2024-08-31 alle 11 55 03](https://github.com/user-attachments/assets/381080c0-b8ee-4cbb-a268-7a25d00c2c25)
![Schermata 2024-08-31 alle 11 54 59](https://github.com/user-attachments/assets/783fce52-997b-4578-bad6-31b3cc6b3523)

**Details:**
 - Device: MacBook Pro 15-inch, Mid 2015
 - macOS: 12.7.6
 - Application version: 2.11.7

PR #2296, opened 2024-12-31
Aggregate RAM usage by parent process

andrewr-dev:
Proposal to fix the following with a new Combine processes option:
Fixes: https://github.com/exelban/stats/issues/2099
Fixes: https://github.com/exelban/stats/issues/2162


exelban:
hi. Thanks for your contribution, at first look it looks fine. But I will need some time to check how it works and if all cases are covered. Will back with feedback later.

PS: please ping me in case I will not provide any feedback in 7 days

exelban:
Zrzut ekranu 2025-01-11 o 18 21 43

something is totally off

andrewr-dev:
Code was displaying resident_size
Changing it to display phys_footprint seems like a closer match to the Memory column in Activity Monitor

exelban:
Hi, are you at least run and check this PR?
The second time it's totally of from what activity monitor shows.

Zrzut ekranu 2025-01-19 o 12 54 28

andrewr-dev:
The previous screenshot is showing individual process usage. You need to look at the process hierarchy, eg. lldb-rpc-server is a child process of XCode so the memory usage is added together. Similarly, all the Safari processes are added together and grouped under Safari. 
The InferenceProviderService process is owned by another user (_modelmanagerd) and the memory usage for this process not shown.

andrewr-dev:
Also I changed a line in Modules/RAM/popup.swift to correct an existing issue with the way GB is displayed (changed GB RAM to use powers of 2 rather than decimal). The screenshot seems to be missing this minor correction.

exelban:
Hi. Sorry for so long response. There is still some problems with non combined processes. It does not show all processes for some reason: Activity Monitor - current reader - this PR reader with disabled "combine processes"

Zrzut ekranu 2025-03-16 o 11 36 35

For some reason this implementation missing some processes

exelban:
This implementation works fine when combined processes is enabled. Maybe will be better to separate these modes:
- current implementation when combined processes is disabled
- your implementation for enabled combined processes option

PR #2333, opened 2025-01-13
fix bundle version

luckman212:
2.11.25 hotfix release still contained previous version string 2.11.24

image


exelban:
oh, that's what happened when I trying to do something quick(

exelban:
thx, but its not just a change in the file. I need to rebuild a release

exelban:
maybe it's a good moment to automate the build process

litori:
Updating the release after it has been released screws up all the homebrew installs.  It continues to fail the checksum test and 2.11.25 no longer installs on anyone running homebrew.  Please use another version number when updating releases, otherwise no one on homebrew can install it.

PR #2353, opened 2025-01-24
lang: added missing translation

fadedflower:


exelban:
hi. I have take a quick look and there is no string like "Normal" in the project

fadedflower:
It is in the memory pressure notification setting

exelban:
ok, but just adding a string to the language file does not mean that it will be translated. Again, there is no such string in the project. Memory pressures are not translated since its system value. So if you want to add translation to that you need to add according code for that to be translated.

fadedflower:
@exelban It IS translated.
SCR-20250124-rans

fadedflower:
It is translated by calling selectView in Kit/extensions.swift
```swift
func selectView(action: Selector, items: [KeyValue_p], selected: String) -> NSPopUpButton {
        let select: NSPopUpButton = NSPopUpButton(frame: NSRect(x: 0, y: 0, width: 50, height: 28))
        select.target = self
        select.action = action
        
        let menu = NSMenu()
        items.forEach { (item) in
            if item.key.contains("separator") {
                menu.addItem(NSMenuItem.separator())
            } else {
                let interfaceMenu = NSMenuItem(title: localizedString(item.value), action: nil, keyEquivalent: "")
                interfaceMenu.representedObject = item.key
                menu.addItem(interfaceMenu)
                if selected == item.key {
                    interfaceMenu.state = .on
                }
            }
        }
        select.menu = menu
        
        return select
    }
```

exelban:
hmm, yeap. You are right. Missed that, was looking only for strings. So LGTM then, tnx

PR #2361, opened 2025-01-26
lang: added missing translation

fadedflower:


exelban:
why do you create a new PR?

fadedflower:
The string "Toggle the module" is not translated. So I closed the previous one and created a new PR.

exelban:
But why you not add a commit? Please do not spam with PR, use commits if you want to make changes

fadedflower:
I see. Sorry about that.

PR #2447, opened 2025-03-16
Support named timezones

cleishm:
Instead of just UTC+/UTC- offsets, support named timezones. These will adjust for DST correctly.

### DIFF
--- a/Kit/extensions.swift
+++ b/Kit/extensions.swift
@@ -540,7 +540,7 @@ public extension Data {
 }
 
 public extension Date {
-    func convertToTimeZone(_ timeZone: TimeZone) -> Date {
+    func adjustToTimeZone(_ timeZone: TimeZone) -> Date {
         return addingTimeInterval(TimeInterval(timeZone.secondsFromGMT(for: self) - TimeZone.current.secondsFromGMT(for: self)))
     }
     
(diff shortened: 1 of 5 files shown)


exelban:
Hi. Basically LGTM. I would like to merge that now, but before one feature is missed - support of current times. It means that if I merge that version it will remove all clocks the user has. I see 2 options: you can add that or I will merge that next week and will make some workaround.

PS: also the select box needs to be adjusted to correctly fit the name:

Zrzut ekranu 2025-03-16 o 11 31 08

cleishm:
I think I've resolved those concerns, although I'm not especially familiar with the UI frameworks. I think that control would also be improved if it was replaced with a text field using autocomplete and validation (rather than a very, very long list), but that seems a little too ambitious for now.

Kit/constants.swift
```diff
@@ -22,7 +22,7 @@ public struct Popup_c_s {
 }
 
 public struct Settings_c_s {
-    public let width: CGFloat = 540
+    public let width: CGFloat = 684
```
exelban, line 25: why did you change the settings window?

cleishm: Widening the Timezone column made the Name column unacceptably small, as the table is sized to fit the settings window. Increasing the overall width fixed that. But perhaps there is a better approach?

exelban: if I will increase the size of the settings window for such a small feature the width will be 10000+.
- scrollable table
- resizing another tabs
- etc

but the best solution will be just crop the text inside the select with `...`

cleishm: Whatever you think best. Just be aware that timezones are most significant on the right hand side, so truncating may make them ambiguous. Maybe a text field with autocomplete (and validation) for entry would be better, then you can truncate to the left when displaying.

Issue #2515, opened 2025-04-15
[Feature] Disk Free Space in GB/TB


romanr:
To have plain free space without total and not in percent free space.  
There's Free space widget but it displays total and free space is too small to read at a glance. as half as small as it could be.   
I don't know who needs to know total, it's always same number.  

(And there's Free space in percent. To know how much is 1% or 3% - you have to do mental mathematics to make use of this number.)

exelban:
what do you mean `displays total and free space` ?
how can you visualize the total space of the disk in the menu bar?

romanr:
I didn't know all the terminology of the project, sorry.  I mean Mini widget. Here is the prototype:

![Image](https://github.com/user-attachments/assets/d0ab13d7-3e45-43e6-8de4-94e37b4cc60c)

romanr:
I can implement it myself.


By the way, The "Colour - based on utilisation"

![Image](https://github.com/user-attachments/assets/25e6fbeb-8525-4baa-b0db-9ccc031e5a16)

Does not apply to modern disks when utilisation measured in percent. When disks were small, yes. But now with terabyte sized disks, 5% of 1TB  may not necessarily be critical displayed in red. it's 50GB. While in the past, 5% of 10GB disk free is alarming.
Configurable threshold like in iStat would work here.

PR #2532, opened 2025-04-30
Mini Widget - added support for absolute units (TB/GB/MB) displaying the Disk free space

romanr:
Hello,
Thank you for this project, it's fantastic!

- Added switch in Settings - Disk - Mini Widget.
- added formatting with character limit for compact Mini widget
- displaying absolute value with resizing in Widget.

Issue #2515

CleanShot-CleanShot-04301924@2x

Next I'm thinking to implement editing of thresholds for widget colour changes (e.g. red is low memory or disk)

### DIFF
--- a/Kit/Widgets/Mini.swift
+++ b/Kit/Widgets/Mini.swift
@@ -52,6 +52,12 @@ public class Mini: WidgetWrapper {
                     if let value = configuration["Value"] as? String {
                         self._value = Double(value) ?? 0
                     }
+                    if let freeDiskSize = configuration["FreeDiskSize"] as? Int64 {
+                        self._usedBytes = freeDiskSize
+                    }
+                    if let absoluteUnits = configuration["AbsoluteUnits"] as? Bool {
+                        self.absoluteUnitsState = absoluteUnits
+                    }
                 }
             }
             
@@ -69,7 +75,7 @@ public class Mini: WidgetWrapper {
                     self.colorState = defaultColor
                 }
             }
-            if let configSupportsGB = config?["SupportsGB"] as? Bool {
+            if let configSupportsGB = configuration["SupportsGB"] as? Bool {
                 self.supportsGB = configSupportsGB
             }
         }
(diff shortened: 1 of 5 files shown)


exelban:
hi, sorry for the long response. The problems I have with this PR:
- You change the logic of the widget: the mini widget was designed to show only the percentage value. And it will not change. I don't like a hack in the way: ok I can just add another if and it will work. Cause tomorrow someone will come and ask for a trigger: free/used space, after total..... etc. 
- I don't see any reason to see an absolute value of the disk in the menu bar, for me percentage is a more understandable way to see how much space left on the disk quickly

And the biggest issue here: there is a widget that basically allows you to show any value you want in the menu bar. It just needs to be enabled for the disk module since it's already enabled in the RAM.

romanr:
> hi, sorry for the long response. The problems I have with this PR:
> 
> * You change the logic of the widget: the mini widget was designed to show only the percentage value. And it will not change. I don't like a hack in the way: ok I can just add another if and it will work. Cause tomorrow someone will come and ask for a trigger: free/used space, after total..... etc.

Believe me, I don't like having to use `if` there either. I checked for similar patterns; if there had been a widget that implements switching of data type, I would have copied that implementation. Maybe you can suggest, I don't want to reinvent anything if you have ideas about how to do it. Someone spent a lot of time understanding the project structure, building, and testing this function. If you want things done in a certain way, you can document that as a guideline. Yet, you just close it instead of returning to fix, as if you disqualify me from making a PR? And I am not sure what "someone will request something else tomorrow" has to do with anything I've done. Anyone will, just like, look into code and seeing no other way implement as they can.

> * I don't see any reason to see an absolute value of the disk in the menu bar, for me percentage is a more understandable way to see how much space left on the disk quickly
>

example from an hour ago; Disk size is 2TB and Free Space widget shows 1.2% Free. I work with large video files and have to to copy 28GB raw video. Is it going to fit 1.2%? I can not calculate that in my mind. If only there was a widget that shows free space in absolute units! 
Maybe you don't have to work with large files, or you are very good at maths, you don't need it. Percentage space made sense 15 years ago when disks were still small relative to file sizes.

> And the biggest issue here: there is a widget that basically allows you to show any value you want in the menu bar. It just needs to be enabled for the disk module since it's already enabled in the RAM.

I am not familiar with the project and no idea what you referring to "there is such  widget", why obfuscate and not tell what is the solution you referring to?

exelban:
Would be better to create an feature request and discuss that before.

leave that to me. I will add this option to the Disk module that this week.
PS: please ping me if I will not provide any information till Saturday.

romanr:
> Would be better to create an feature request and discuss that before.
> 

Which before I did, #2515 and you closed with "not gonna do it".

exelban:
yea, cause you wanted to do that in mini widget. And that will not be done.

PR #2606, opened 2025-06-22
feat: added pin feature for popup

RamanaRaj7:
-A popup can be pinned by pressing the pin icon near the setting icon in the popup, when pinned it will stay on top even while other apps are being accessed 

Screenshot 2025-06-23 at 1 32 28 AM

https://github.com/user-attachments/assets/e72cf599-7bc8-4f24-804b-2fbdc9504015

### DIFF
--- a/Modules/Net/readers.swift
+++ b/Modules/Net/readers.swift
@@ -669,42 +669,26 @@ internal class ConnectivityReader: Reader<Network_Connectivity> {
     
     private var _status: Bool? = nil
     private var status: Bool? {
-        get {
-            self.variablesQueue.sync { self._status }
-        }
-        set {
-            self.variablesQueue.sync { self._status = newValue }
-        }
+        get { self.variablesQueue.sync { self._status } }
+        set { self.variablesQueue.sync { self._status = newValue } }
     }
     
     private var _timeoutTimer: Timer?
     private var timeoutTimer: Timer? {
-        get {
-            self.variablesQueue.sync { self._timeoutTimer }
-        }
-        set {
-            self.variablesQueue.sync { self._timeoutTimer = newValue }
-        }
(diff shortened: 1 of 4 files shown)


exelban:
I don't like this implementation (UI/UX part)

RamanaRaj7:
> I don't like this implementation (UI/UX part)

any suggestions ?
what do you think about the name of the popup being the pin button?

RamanaRaj7:
how about adding a toggle in the settings to enable or disable the pin in the popup? so only the people who want it can enable it. let me know what you think.

exelban:
@RamanaRaj7 yeap, it could work

RamanaRaj7:
> @RamanaRaj7 yeap, it could work

added the toggle

RamanaRaj7:
any other suggestions ?

exelban:
sorry, have no time to look at.
Ok, so looks like we misunderstood each other. I think you are going to add a setting to the settings: `popup always active`. And for sure not adding a toggle to show/hide the pin. I do not like the idea of having a pin in the popup. There are 2 ways:
- or have a pin mechanism that will be very minimalistic in the popup
- or to have that as an option in the settings that keeps the popup always on top

RamanaRaj7:
what about the name of the popup being the pin? 
so no pin icon is there as you expect and when pressed on the name of the popup turns blue signifying the popup is pinned
let me know what you think.

exelban:
no one will ever find this feature

RamanaRaj7:
thats a thing but lets add a toggle for that to in settings so when going through the settings they might find it and try it
and also when hovered over the name let it show some hint like in the below image
Screenshot 2025-07-22 at 1 13 45 AM

PR #2648, opened 2025-07-23
Localize timeseries tooltips

yuzawa-san:
I found the tooltips were not localized, so I formatted the date/time in the user's locale.

example: for locale en_US
image

### DIFF
--- a/Kit/plugins/Charts.swift
+++ b/Kit/plugins/Charts.swift
@@ -73,6 +73,8 @@ internal func scaleValue(scale: Scale = .linear, value: Double, maxValue: Double
     return y
 }
 
+private let subtitleFont = NSFont.systemFont(ofSize: 9, weight: .medium)
+
 private func drawToolTip(_ frame: NSRect, _ point: CGPoint, _ size: CGSize, value: String, subtitle: String? = nil) {
     guard !value.isEmpty else { return }
     
@@ -111,7 +113,7 @@ private func drawToolTip(_ frame: NSRect, _ point: CGPoint, _ size: CGSize, valu
     str.draw(with: rect)
     
     if let subtitle {
-        attributes[NSAttributedString.Key.font] = NSFont.systemFont(ofSize: 9, weight: .medium)
+        attributes[NSAttributedString.Key.font] = subtitleFont
         attributes[NSAttributedString.Key.foregroundColor] = (isDarkMode ? NSColor.white : NSColor.textColor).withAlphaComponent(0.7)
         rect = CGRect(x: position.x, y: position.y, width: size.width-8, height: 9)
         str = NSAttributedString.init(string: subtitle, attributes: attributes)
@@ -152,7 +154,9 @@ public class LineChartView: NSView {
         
         super.init(frame: frame)
         
-        self.dateFormatter.dateFormat = "dd/MM HH:mm:ss"
+        self.dateFormatter.locale = Locale.current
+        self.dateFormatter.dateStyle = .short
+        self.dateFormatter.timeStyle = .medium
         
         self.addTrackingArea(NSTrackingArea(
             rect: CGRect(x: 0, y: 0, width: self.frame.width, height: self.frame.height),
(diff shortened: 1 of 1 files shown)


exelban:
LGTM, thx

exelban:
@yuzawa-san no monospaced font, it causes crashes on some macs

yuzawa-san:
i updated it.

do you know which version(s) it is an issue? i think this app could benefit nicely from monospaced digit system font being used more widely especially given how many digits we display. would it be worthwhile to add (not in this PR) some sort of polyfill like this
```
extension NSFont {
func safeMonospacedDigitSystemFont(... ){
 if #available(macOS 13.0, *)  {
   return NSFont.monospacedDigitSystemFont(...)
 } else {
  return NSFont.SystemFont(...)
 }
 }
}
```
i can make an issue and/or give it a try if you want

exelban:
no monospaced font. I was ready to merge that, but I have noticed that you make a lot of unnecessary changes. What is for?
Im pretty sure that static width was for some reason since there was some problem with dynamic width.

It's a PR for localize timeseries tooltips, so please make sure it's only one change.

yuzawa-san:
reverted back to the initial commit you were ready to approve

Kit/plugins/Charts.swift
```diff
@@ -327,10 +331,12 @@ public class LineChartView: NSView {
                 path.stroke()
                 
                 let date = self.dateFormatter.string(from: nearest.value.ts)
+                let subtitleWidth = date.widthOfString(usingFont: subtitleFont)
+                let width = max(70, subtitleWidth) + 8
                 let roundedValue = (nearest.value.value * 100).rounded(toPlaces: 2)
                 let strValue = roundedValue >= 1 ? "\(Int(roundedValue))\(suffix)" : "\(roundedValue)\(suffix)"
                 let value = toolTipFunc != nil ? toolTipFunc!(nearest.value) : strValue
-                drawToolTip(self.frame, CGPoint(x: nearest.point.x+4, y: nearest.point.y+4), CGSize(width: 78, height: height), value: value, subtitle: date)
+                drawToolTip(self.frame, CGPoint(x: nearest.point.x+4, y: nearest.point.y+4), CGSize(width: width, height: height), value: value, subtitle: date)
```
exelban, line 339: how this change is related to the date format?

yuzawa-san: it is because not every locale has the same format and so different locales have different width resulting strings. if i did not add this, some locales would have their strings truncated based on the existing width.

exelban: there was a reason why I do that with static width: with dynamic there was too many resizes and tooltip box "jumps" all the time

yuzawa-san: ha thats why i tried the monospaced font, before you flagged that as not working. once again it would _really_ awesome to figure out what version that is broken in and gracefully fall back, but that is out of the scope of this PR. i just retested this. here is a video in my locale. typically only a single digit is changing. is this acceptable? 

https://github.com/user-attachments/assets/f4eafe7a-6f9a-4ec1-916d-b5ef86f92f21

exelban: nope, I really do not like jumping boxes. If you want to get more info about monospaced font just search in issues about that, there was a few times when I use that font.

yuzawa-san: given the variance is within a few pixels in the absence of a monospace font. can i just calculate the width for the locale in the constructor add like a few pixels of buffer and then reuse that width in the render to prevent the jitter?

yuzawa-san:
this is all becoming too much trouble, so i propose https://github.com/exelban/stats/pull/2689 instead

yuzawa-san:
@exelban i have attempted to fix the jitter on this. the solution i came up with was to keep a max width of the localized date string. so it will converge to the longest value it encounters and hold onto it.

yuzawa-san:
why?

yuzawa-san:
why?

PR #2689, opened 2025-09-01
Use ISO date format for timeseries tooltips

yuzawa-san:
- replacement for https://github.com/exelban/stats/pull/2648
- i decided to select a globally recognizable (and fixed width) time format instead of a specific locale's format

image

### DIFF
--- a/Kit/plugins/Charts.swift
+++ b/Kit/plugins/Charts.swift
@@ -152,7 +152,7 @@ public class LineChartView: NSView {
         
         super.init(frame: frame)
         
-        self.dateFormatter.dateFormat = "dd/MM HH:mm:ss"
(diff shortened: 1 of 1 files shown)


yuzawa-san:
why?

exelban:
This format is not readable for me.

yuzawa-san:
thanks for the response. it is bit jarring to have things seemingly randomly closed for no reason.

could you please elaborate? is something broken or the rendering off? please share a screenshot if so.

exelban:
this format is not readable for me, screenshot of what you expect?

I will not accept a change that will make something unusable for me. Previous PR with local ts was ok, but for some reason you closed it.

PR #2719, opened 2025-09-22
Fix: WiFi SSID retrieval for macOS 15.6+ compatibility

RamanaRaj7:
Recent macOS versions (15.7+) redacted SSID details from `system_profiler', breaking the previous Wi-Fi SSID retrieval method. 

redacted

A workaround for this is to use  ` en="$(networksetup -listallhardwareports | awk '/Wi-Fi|AirPort/{getline; print $NF}')"; ipconfig getsummary "$en" | grep -Fxq "  Active : FALSE" || networksetup -listpreferredwirelessnetworks "$en" | sed -n '2s/^\t//p' ` which is implemented here.

This was also tested with macOS 26.

if you don't see it working after Build is succeed try using cleanmymac or some other cleaning apps to delete stats completely and Build it again.

### DIFF
--- a/Kit/module/settings.swift
+++ b/Kit/module/settings.swift
@@ -73,7 +73,7 @@ open class Settings: NSStackView, Settings_p {
         self.distribution = .fill
         self.spacing = Constants.Settings.margin
         self.edgeInsets = NSEdgeInsets(
-            top: Constants.Settings.margin,
+            top: 0,
             left: Constants.Settings.margin,
             bottom: Constants.Settings.margin,
             right: Constants.Settings.margin
@@ -312,12 +312,18 @@ private class WidgetSelectorView: NSStackView {
     private var background: NSVisualEffectView = {
         let view = NSVisualEffectView(frame: NSRect.zero)
         view.blendingMode = .withinWindow
-        view.material = .contentBackground
+        if #available(macOS 26.0, *) {
+            view.material = .titlebar
+        } else {
+            view.material = .contentBackground
+        }
         view.state = .active
         view.wantsLayer = true
         view.layer?.cornerRadius = 5
         return view
     }()
+    
+    private var separator: NSView?
     
     fileprivate init(module: String, widgets: [SWidget], stateCallback: @escaping () -> Void) {
         self.module = module
@@ -366,8 +372,9 @@ private class WidgetSelectorView: NSStackView {
         let separator = NSView()
         separator.identifier = NSUserInterfaceItemIdentifier(rawValue: "separator")
         separator.wantsLayer = true
-        separator.layer?.backgroundColor = NSColor(red: 213/255, green: 213/255, blue: 213/255, alpha: 1).cgColor
+        separator.layer?.backgroundColor = NSColor.separatorColor.withAlphaComponent(isDarkMode ? 0.35 : 0.15).cgColor
         self.addArrangedSubview(separator)
+        self.separator = separator
         
         inactive.forEach { (widget: WidgetPreview) in
             self.addArrangedSubview(widget)
@@ -379,7 +386,7 @@ private class WidgetSelectorView: NSStackView {
         NSLayoutConstraint.activate([
             self.heightAnchor.constraint(equalToConstant: Constants.Widget.height + (Constants.Settings.margin*2)),
             separator.widthAnchor.constraint(equalToConstant: 1),
-            separator.heightAnchor.constraint(equalTo: self.heightAnchor, constant: -6)
+            separator.heightAnchor.constraint(equalTo: self.heightAnchor, constant: -18)
         ])
     }
     
@@ -389,6 +396,7 @@ private class WidgetSelectorView: NSStackView {
     
     override func updateLayer() {
         self.background.setFrameSize(self.frame.size)
+        self.separator?.layer?.backgroundColor = NSColor.separatorColor.withAlphaComponent(isDarkMode ? 0.35 : 0.15).cgColor
     }
     
     override func mouseUp(with event: NSEvent) {
--- a/Stats/Views/AppSettings.swift
+++ b/Stats/Views/AppSettings.swift
@@ -65,7 +65,7 @@ class ApplicationSettings: NSStackView {
         
         let scrollView = ScrollableStackView(orientation: .vertical)
         scrollView.stackView.edgeInsets = NSEdgeInsets(
-            top: Constants.Settings.margin,
+            top: 0,
             left: Constants.Settings.margin,
             bottom: Constants.Settings.margin,
             right: Constants.Settings.margin
--- a/Stats/Views/Dashboard.swift
+++ b/Stats/Views/Dashboard.swift
@@ -181,7 +181,7 @@ class Dashboard: NSStackView {
         
         let scrollView = ScrollableStackView(orientation: .vertical)
         scrollView.stackView.edgeInsets = NSEdgeInsets(
-            top: Constants.Settings.margin,
+            top: 0,
             left: Constants.Settings.margin,
             bottom: Constants.Settings.margin,
             right: Constants.Settings.margin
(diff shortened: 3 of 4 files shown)


exelban:
It does not work properly. I'm open to the discussion, but only if you understand the solution you provide and understand in which cases it simply does not work. But even worse, it returns the wrong value.

If you just copy the first solution from Stack Overflow, there is nothing to discuss.

RamanaRaj7:
> It does not work properly. I'm open to the discussion, but only if you understand the solution you provide and understand in which cases it simply does not work. But even worse, it returns the wrong value.
> 
> If you just copy the first solution from Stack Overflow, there is nothing to discuss.

I got the solution from this [https://stackoverflow.com/questions/78994709/how-to-get-the-current-wi-fi-ssid-in-swift-on-macos](url) stack overflow thread .

Well this basically works by getting the list of preferred networks on the connected network interface and getting the top name from that list since the most recently connected network shows up first here.

As I have mentioned in the first comment `if you don't see it working after Build is succeed try using cleanmymac or some other cleaning apps to delete stats completely and Build it again.`  After a clean installation it works fine (meaning you have not installed the app before or you don't have any persistent data of the application) but when you just simply build the app without deleting the application without persistent data it doesn't work. I'm currently looking on a fix for this.

I'm not sure how it returns wrong value . if possible can you let me know more info on this ?

fiwswe:
@RamanaRaj7 Note: I don't use Stats, but I noticed your PR.

The order of SSIDs in the output of `networksetup -listpreferredwirelessnetworks` is undocumented to my knowledge. Grabing the first one may not be the correct answer.

In my case this list contains networks I have configured on my Mac, as well as those my iPhone knows about. In fact the first in my list is the SSID at a remote location where my Mac has never been.

---
Also in the command line version in the comment, detecting deactivated interfaces should be followed by `&&` not `||`. I.e.:
`en="$(networksetup -listallhardwareports | awk '/Wi-Fi|AirPort/{getline; print $NF}')"; ipconfig getsummary "$en" | grep -Fxq "  Active : FALSE" && networksetup -listpreferredwirelessnetworks "$en" | sed -n '2s/^\t//p'`
That at least works better than your version on my Mac where Wi-Fi on but not connected to any SSID (enabling AirDrop without having a second redundant connection to the LAN). Your actual code seems to do this correctly if I read it correctly.

RamanaRaj7:
@fiwswe you might be right `The order of SSIDs in the output of networksetup -listpreferredwirelessnetworks is undocumented to my knowledge. Grabing the first one may not be the correct answer.` about this is not documented , but I tested it on my Mac where the results showed the most recently connected network on the top of the list.

`In my case this list contains networks I have configured on my Mac, as well as those my iPhone knows about. In fact the first in my list is the SSID at a remote location where my Mac has never been.` well I not sure why this happens though , can you provide details on which kind of Mac you use and its macOS version? have you connected your iPhone to the network on the remote location you mentioned or any other device which share your iCloud account?

`Also in the command line version in the comment, detecting deactivated interfaces should be followed by && not ||.` Here the goal is to get the list of wifi network when the network is active alone ,Having && makes it show preferred networks if wifi is inactive.

fiwswe:
@RamanaRaj7 
> can you provide details on which kind of Mac you use and its macOS version?

On this machine (Mac mini M2 Pro) macOS 14.8 Sonoma is currently running. And before you say it, yes this might be different on more recent macOS versions. But that would only prove my point that you can't rely on the order.

> have you connected your iPhone to the network on the remote location you mentioned or any other device which share your iCloud account?

Yes. The list of SSIDs returned by `networksetup -listpreferredwirelessnetworks` is a union of the sets of SSIDs the Mac has connected to and those iPhones and other devices in the Apple Account have connected to.

The point is, unless you can cite official documentation that states that the currently active SSID is listed first, you can't rely on that being the case. And going by observed output if you only tested this on very few machines is no guarantee either. Even if that happens to work during your tests, Apple could change the behaviour with the next update.

Additionally my use case with Wi-Fi on and no connected SSID, while not that common, is certainly not handled correctly by your code. You assume that Wi-Fi on means there is a current SSID.

> Here the goal is to get the list of wifi network when the network is active alone ,Having && makes it show preferred networks if wifi is inactive.

I only did a quick test of your shell code. In your original version the output was the aforementioned SSID of a remote network the Mac had never seen itself. In my corrected version the output was empty which seems more correct. In both cases the Mac was configured with Wi-Fi on but no connected Wi-Fi network, because I'm using an Ethernet connection and having both caused some issues, e.g. with maintaining SSH sessions. This configuration does allow AirDrop to work though, which I occasionally use.

But this is only a side issue.

RamanaRaj7:
>On this machine (Mac mini M2 Pro) macOS 14.8 Sonoma is currently running. And before you say it, yes this might be different on more recent macOS versions. But that would only prove my point that you can't rely on the order.

@fiwswe Thanks for letting me know

>I'm using an Ethernet connection 

But this PR handles only wifi interface not ethernet . Ethernet is handled separately using a different function in the application so this logic doesn’t apply to Ethernet interfaces.

If possible can you connect your Mac mini to your mobile hotspot or wifi and run the command and share the result?

exelban:
There are a few problems with this solution:
- first and the most important: it returns the wrong network name on my MacBook Pro
- this command has so many internal tools that it creates too many places that could be broken, and it will cause more problems in future debugging.
- Apple tries so hard to prevent getting ssid that I'm pretty sure this way will be blocked in the near future as well

There is one good way to do that: to have location permission for the app. In that case all tools will return ssid without any problem. But also I will not add this permission to be on by default or even ask the user for that.
I will try to make a workaround, but only when I find time to make more changes in the Network popup window.

fiwswe:
> If possible can you connect your Mac mini to your mobile hotspot or wifi and run the command and share the result?

The connected network popped to the top of the list (and stayed there after disconnecting). Details:
* Start: Wi-Fi on, not connected to any network. First listed SSID is from a remote location that the Mac has never seen. The local SSID is present in the list.
* Connected the Mac to the local SSID.
* Ran the command (`networksetup -listpreferredwirelessnetworks en1`). The local SSID is now first in the list.
* Set the Wi-Fi network to **not** reconnect automatically.
* Turned Wi-Fi off and on again. Wi-Fi is now on and not connected to any network again.
* Ran the command again. The local SSID remains at the top of the list.

Conclusion: The observed behaviour seems like the last used SSID of all devices using the same Apple Account is listed first. However that is not quite correct either. My iPhone is presently connected to the above mentioned local SSID. So I did another test:
* Switched my iPhone to a second local SSID.
* Ran the command on the Mac. Now this second SSID is first in the list.
* Switched my iPhone back to to the first local SSID.
* Ran the command on the Mac. Now the first local SSID is first in the list again.

So the last SSID any device in the Apple Account connected to is listed first may be partially correct. Note that the remote SSID I mentioned was used by the iPhone a few weeks ago. But of course when I came back home, it reconnected to the first local SSID. And that event did not trigger any reordering of the list apparently. Seems inconsistent to me.

To me all of this means that the first SSID in the list can not be relied on to be the SSID the Mac is connected to.

RamanaRaj7:
@fiwswe Thanks for putting a great effort on this! really appreciate it.
>To me all of this means that the first SSID in the list can not be relied on to be the SSID the Mac is connected to.

noted 🫡

fjh658:
@exelban 
```
// Pull SSID from saved profiles (public API; no scanning). Useful when cw.ssid() is redacted.
private static func ssidFromProfiles(_ cw: CWInterface) -> String? {
    guard let cfg = cw.configuration(),
            let set = (cfg.value(forKey: "networkProfiles") as? NSOrderedSet),
            let first = set.firstObject as? CWNetworkProfile,
            let raw = first.ssid, !raw.isEmpty else {
        return nil
    }
    // Normalize smart quotes and whitespace to match our other sources.
    return raw.replacingOccurrences(of: "’", with: "'")
                .replacingOccurrences(of: "‘", with: "'")
                .trimmingCharacters(in: .whitespacesAndNewlines)
}

OR

private static func ssidFromProfiles(_ cw: CWInterface) -> String? {
    guard let set = cw.configuration()?.networkProfiles else { return nil }
    for case let p as CWNetworkProfile in set {
        if let s = p.ssid, !s.isEmpty { return s }
    }
    return nil
}

if ssid == nil {
    ssid = ssidFromProfiles(cw)
}
```

exelban:
@fjh658 thank you!

Issue #2637, opened 2025-07-17
Include previous IP in IP changed notification


PuzzledUser:
It'd be helpful to see the previous (public or local) IP in the notification to better understand the change.

<img width="376" height="86" alt="Image" src="https://github.com/user-attachments/assets/ba23b4a2-e9fc-4d3e-bbbd-0e103a6ba9f5" />

i.e. Public IP changed from %1

PuzzledUser:
Please also add the IP to the Private IP changed notification

TangilHossain:
I would Like to work on it. If no one is assigned on it yet. @exelban

TangilHossain:
Fixed it and will be doing a PR for review.

<img width="344" height="225" alt="Image" src="https://github.com/user-attachments/assets/e4af5df8-353c-45e7-9ed4-15f6d0f58612" />

PuzzledUser:
I see the notification now shows New IP: xxx.xxx.xxx.xxx. This doesn't give any clue as to what it was before. The current IP is available in the Network menu, so this implementation adds no new information. `Public IP changed from %1` would help.

PR #2728, opened 2025-09-29
Solve to issue ID: #2637

TangilHossain:
Added details on ip change notification. 
A little change on the structure of the notifications file.


exelban:
hi, thanks for PR. Basically looks fine, only one thing need to be fixed: you don't use translation for this notification

TangilHossain:
Updated the strings files. Hope it matches your expectations.
image

TangilHossain:
@exelban please check.

exelban:
please remove the translations, I will add it by myself, or make that in a separate commit. 
Also there is a total mess with the project configuration (do not change project configuration)

also, again, please use the way Stats is handling the translations (localizedString func)

PR #2745, opened 2025-10-05
feat: Add Ports module for monitoring active network ports

dogukanakin:
- Add new Ports module to monitor active TCP ports
- Display port number, process name, and PID
- Add kill functionality with confirmation dialog
- Support for Turkish and English localization
- Add ModuleType.Ports to constants
- Include popup UI with port list and kill buttons
- Add settings page for update interval configuration

This feature allows users to:
- View all active TCP ports on their system
- See which processes are using each port
- Kill processes safely with confirmation dialog
- Refresh port list manually

This feature will be particularly useful for developers and system administrators who need to monitor and manage network ports on their Mac.

### DIFF
--- a/Kit/constants.swift
+++ b/Kit/constants.swift
@@ -66,6 +66,7 @@ public enum ModuleType: Int {
     case battery
     case bluetooth
     case clock
+    case Ports
     
     case combined
     
@@ -80,6 +81,7 @@ public enum ModuleType: Int {
         case .battery: return "Battery"
         case .bluetooth: return "Bluetooth"
         case .clock: return "Clock"
+        case .Ports: return "Ports"
         case .combined: return ""
         }
     }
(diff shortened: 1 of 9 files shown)


exelban:
Hi, thx for your PR. There are a few points to this:
- first of all there will be no Ports module in the app for sure
- in theory it could be a part of the Network module (since it's a part of the network)
- the last part: I'm not sure it fits the app, maybe it's a nice feature that could be added to the app as an option to see in the popup module. But it's quite a niche feature, but I think it could be added to the app in some way.

Also, why only TCP? UDP is quite popular for some tools.

dogukanakin:
Hi, thank you for the feedback!

You're absolutely right on all points:

1. **Standalone module**: I agree it should be integrated into the Network module instead of being separate
2. **TCP only**: Good catch! Should definitely include UDP as well
3. **Niche feature**: Making it an optional tab within Network module makes perfect sense

I understand this approach would be much better - integrating ports monitoring as an optional feature within the existing Network module rather than creating a whole new module.

Since I'm not planning to implement these changes right now, feel free to close this PR. The code is there if you or other contributors want to use it as a reference for implementing this feature in the Network module in the future.

Thanks for considering the feature!

exelban:
I'm planning for a long time to have some monitor features in the window app. And I think it would fit nicely there. Thank you for the code base, i'm sure that it will be helpful.

Issue #2759, opened 2025-10-15
Incorrect SMART read/write data


xiaoran007:
**Describe the bug**
In recent versions of stats, the Disk module displays incorrect values for SMART read/write data. The numbers appear to be inconsistent with system-level statistics (for example `smartctl`).

Here is the result reported by `smartctl`:
```text
SMART/Health Information (NVMe Log 0x02, NSID 0xffffffff)
Critical Warning:                   0x00
Temperature:                        32 Celsius
Available Spare:                    100%
Available Spare Threshold:          99%
Percentage Used:                    3%
Data Units Read:                    65,869,637 [33.7 TB]
Data Units Written:                 79,576,087 [40.7 TB]
Host Read Commands:                 1,902,852,045
Host Write Commands:                1,211,702,099
Controller Busy Time:               0
Power Cycles:                       159
Power On Hours:                     1,320
Unsafe Shutdowns:                   5
Media and Data Integrity Errors:    0
Error Information Log Entries:      0
```
But the Read / Write count reported by `stats` is only `111 MB`.

<img width="267" height="81" alt="Image" src="https://github.com/user-attachments/assets/1fc3224a-5f05-41cd-aff7-c486e4ce73bc" />

I found this issue both on my M1 Mac and M4 Mac.

**Details:**
 - Device: Mac mini (M4)
 - macOS: 15.6.1
 - Application version: 2.11.57

xiaoran007:
I have fixed this issue and created a PR #2760  and I built and tested it on my M4 Mac, and it's now working. Hope this helps.

PR #2760, opened 2025-10-15
Fix: Correct NVMe SMART data units read/written calculation

xiaoran007:
The disk SMART module had one critical issue #2759 

*  **Incorrect SMART data parsing** - only reading the first byte (instead of all 16 bytes) of NVMe SMART `data_units_read` and `data_units_written` fields

**Testing & Verification**
**Before Fix:**
Total read: ~50 MB → resets at ~124 MB
Total written: ~60 MB → resets at ~124 MB
Values meaningless for any drive with significant usage
**After Fix:**
Total read: 33.7 TB ✓
Total written: 40.7 TB ✓
Values accurately match system tools
Verification with smartctl:
```text
Data Units Read:                    65,878,685 [33.7 TB]
Data Units Written:                 79,581,408 [40.7 TB]
```
Result from fixed stats:
截屏2025-10-15 15 32 34

**Impact**
This PR only changes data parsing code, it doesn't impact any other module.

### DIFF
--- a/Modules/Disk/header.h
+++ b/Modules/Disk/header.h
@@ -10,8 +10,6 @@
 //
 
 #include <libproc.h>
-#include <IOKit/IOKitLib.h>
-#include <IOKit/IOCFPlugIn.h>
 #include <IOKit/hidsystem/IOHIDEventSystemClient.h>
 
 struct nvme_smart_log {


Modules/Disk/header.h
```diff
@@ -10,6 +10,8 @@
 //
 
 #include <libproc.h>
+#include <IOKit/IOKitLib.h>
+#include <IOKit/IOCFPlugIn.h>
```
exelban: hi, what this import for?

xiaoran007:
Hi, I have reverted the header changes in `Disk/header.h` to keep this PR minimal and focused only 
on fixing the SMART data reading bug.

PR #2800, opened 2025-11-05
feat: add country flag widget for network module

depapp:
Screenshot 2025-11-05 at 15 08 01
Screenshot 2025-11-05 at 15 16 12
Screenshot 2025-11-05 at 15 07 27

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,4 +18,16 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - run: xcodebuild -scheme Stats -destination 'platform=macOS' -configuration Release archive CODE_SIGNING_ALLOWED=NO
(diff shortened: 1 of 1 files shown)


depapp:
do you have any explanation why my PR closed? @exelban

exelban:
Hi, I believe my previous reply is relevant to this discussion before this PR. There are a couple of reasons why I just close this PR:

- A new widget is not necessary. Stats already have a text widget that can display this information, and adding a configuration flag to that widget should be sufficient.
- Maintenance concerns. I’m not going to add a new widget solely because someone wants a slightly different display. Any new widget requires ongoing maintenance and support, and Stats already have enough widgets to cover most data visualization needs. If I start adding new widgets for every requested variation, we’ll end up with dozens of custom widgets, which would create a significant long-term maintenance burden.

PR #2859, opened 2025-12-13
Update fan control

mihaidimoiu:
Hello!

I have tried to upgrade fan control module and add some improvements.

Now it should do the following: 

### 5. Tried to add support for single fan and 3+ fans 

All modifications are tested on local build using MacOS 26.2 and XCode 26.2.


exelban:
1. No, fan should not be allowed to 0
2. Remove all notifications
3. Remove all new logs

```
After wake, the system should be cool so keep fans in automatic mode
        // User can manually adjust if needed - no need to restore previous settings
```
It's a joke? Why you removing the feature that is in the app and decide that is not needed?

I will not merge that PR for 100%. Cause it has a lot of strange changes. If you know how to fix something that does not work properly related to the fan control please create a new PR where only that will be.

PR #2864, opened 2025-12-15
fix(Clock): implement local caching for settings to prevent UI lag (#2840)

### LINKED ISSUE #2840: Setting Timeformat
**Describe the bug**
If you want to set the time format you have to activate the battery-model otherwise the option is not shown

**Details:**
 - Device: MBP M3
 - macOS: 26.1
 - Application version: 2.11.61

Image

Image

biplavbarua:
### Context
Fixes #2840.

### Problem
The [list](cci:7://file:///Users/biplavbarua/Developer/stats/exportOptions.plist:0:0-0:0) property in [Clock/settings.swift](cci:7://file:///Users/biplavbarua/Developer/stats/Modules/Clock/settings.swift:0:0-0:0) was reading/writing to UserDefaults (JSON encode/decode) on every keystroke in the settings UI. This caused performance issues and prevented users from typing custom formats smoothly.

### Solution
- Implemented a local `cachedList` in `Settings`.
- [init()](cci:1://file:///Users/biplavbarua/Developer/brew_fork/Library/Homebrew/cask/dsl.rb:127:4-178:7) : Loads from Store once.
- `get` : Reads from memory (instant).
- [set](cci:1://file:///Users/biplavbarua/Developer/brew_fork/Library/Homebrew/cask/dsl.rb:225:4-244:7) : Updates memory and persists to Store.

### DIFF
--- a/Modules/Clock/settings.swift
+++ b/Modules/Clock/settings.swift
@@ -20,17 +20,14 @@ let statusColumnID = NSUserInterfaceItemIdentifier(rawValue: "status")
 internal class Settings: NSStackView, Settings_v, NSTableViewDelegate, NSTableViewDataSource, NSTextFieldDelegate {
     public var callback: (() -> Void) = {}
     
+    private var cachedList: [Clock_t] = []
     private var list: [Clock_t] {
         get {
-            if let objects = Store.shared.data(key: "\(self.title)_list") {
-                let decoder = JSONDecoder()
-                if let objectsDecoded = try? decoder.decode(Array.self, from: objects) as [Clock_t] {
-                    return objectsDecoded
-                }
-            }
-            return [Clock.local]
+            return self.cachedList
         }
         set {
+            self.cachedList = newValue
+            
             if newValue.isEmpty {
                 Store.shared.remove("\(self.title)_list")
             } else {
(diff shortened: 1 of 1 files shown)


exelban:
how it fixes #2840?

biplavbarua:
The issue in #2840 (unable to set time format) happens because the previous `list` implementation triggered a full JSON decode & disk read on every single keystroke.
This heavy operation caused the `NSTextField` binding to lag or reset mid-typing, effectively rejecting the user's input before they could finish.
By checking `cachedList` (memory) instead of `Store` (disk) for the getter, the UI binding remains stable and responsive, allowing the user to type the full format string without interruption. The `save()` to disk only happens after the change is committed.

exelban:
it does not, since Store has internal cache, so what you do is basically duplicate the login inside Store. But again, how that fixes #2840?

biplavbarua:
Hi @exelban, thanks for checking this!

You are completely right that `Store` caches the raw `Data`, so we aren't hitting the disk. However, the performance bottleneck here isn't the disk I/O, but the **JSON Decoding**.
Without `cachedList`, the `list` getter calls `JSONDecoder().decode(...)` every single time it's accessed. Since `list` is used by `numberOfRows` and `viewFor` in the TableView, this forces a full JSON decode operation for every frame/cell render, which causes the specific UI stutter/lag reported in #2840 when typing or scrolling.

My change ensures we decode the JSON only once (on init/set) and serve the ready-to-use object from memory thereafter.

exelban:
Oh, yeah, I see the problem with json. But still, how it fixes #2840?
Problem there that settings does not appear because the widget is not active, how cache will solve that?

biplavbarua:
Reviewing the code: `Store` caches the raw `Data`, but `Settings` was previously running `JSONDecoder().decode(...)` on every single access of the `list` property.

Since `list` is accessed frequently by `NSTableView` (numberOfRows, viewFor), this caused massive main-thread blocking (UI Lag).

This PR introduces `cachedList` to store the **decoded** objects, which eliminates the lag. If #2840's reported 'non-appearance' was actually a UI hang/freeze, this fixes it.

exelban:
@biplavbarua I'm pretty sure you do not read #2840. Since it's related to the battery module, not to the clock.

PR #2871, opened 2025-12-20
feat: apply mac material to popup window

fly2xj:
Clipboard_Screenshot_1766224454


exelban:
how it looks on the previous versions of macOS?

fly2xj:
> how it looks on the previous versions of macOS?

This is not a new system-specific feature; it has likely been around for quite some time.

Issue #2789, opened 2025-10-30
Feature Request: Show DNS Servers for IPv4 and IPv6 in Network Module


Ninnamon12:
Currently, the Network module shows public IP and bandwidth stats but not the configured DNS servers.

Request: Add DNS server display for both IPv4 and IPv6 in the Network dropdown, e.g.:

DNS IPv4: 8.8.8.8, 8.8.4.4
DNS IPv6: 2001:4860:4860::8888, 2001:4860:4860::8844

Details:

Pull from system configuration (macOS System Configuration framework).
Update with other network stats.
Show "System default" if not explicitly set.
Optional toggle to keep UI clean.

Would be a helpful addition for troubleshooting and monitoring custom DNS setups.

PR #2875, opened 2025-12-22
Show DNS Servers (IPv4 & IPv6) in Network Module

ukateki:
New feature requested in [#2789](https://github.com/exelban/stats/issues/2789)

### DIFF
--- a/Modules/Net/popup.swift
+++ b/Modules/Net/popup.swift
@@ -475,7 +475,7 @@ internal class Popup: PopupWrapper {
                     
                     if let container = self.addressView {
                         for (idx, dns) in dnsList.enumerated() {
-                            let titleBase = localizedString("DNS servers")
+                            let titleBase = localizedString("DNS")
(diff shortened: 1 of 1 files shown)


ukateki:
Screenshot 2025-12-22 at 10 50 08 AM

exelban:
It should be under details section that could be shown/hidden

exelban:
and for sure not 4, but 2 servers

ukateki:
> It should be under details section that could be shown/hidden

I've already committed the [updates](https://github.com/ukateki/stats/commits/feature/DNS/), moved the DNS rows into the details section, and only show one IPv4 address and one IPv6 address.
Screenshot 2025-12-24 at 4 45 57 PM

But I don't understand what you mean by `that could be shown/hidden`. Should I add a switch to the settings? If yes, where should I add it?

exelban:
this version is ok, but it has no sense to merge that at this point. Since the network windows needs a full redesign and it will be rewritten. And also I will add couple new fields. If there will be no DNS info ping me.

PR #2915, opened 2026-01-14
fix: allow helper connection on codesign error -67062

### LINKED ISSUE #2914: Fan control stopped working after updating to macOS 26.2 (Stats helper XPC fails with -67062)
**Describe the bug**
After upgrading to macOS 26.2 (25C56), fan control stopped working in Stats. Clicking to increase fan speed has no effect. Logs show the app fails to connect to the privileged helper because the codesign check fails with security error `-67062`.

**Details:**
 - Device: MacBook Pro (MacBookPro18,3; Chip: Apple M1 Pro)
 - macOS: 26.2 (25C56)
 - Application version: 2.11.64

arnaldobatista:
Fixes #2914

### DIFF
--- a/SMC/Helper/main.swift
+++ b/SMC/Helper/main.swift
@@ -56,8 +56,11 @@ class Helper: NSObject, NSXPCListenerDelegate, HelperProtocol {
                 return false
             }
         } catch {
-            NSLog("error checking code signing: \(error)")
-            return false
+            let message = String(describing: error)
+            NSLog("error checking code signing: \(message)")
+            if !message.contains("-67062") {
+                return false
+            }
         }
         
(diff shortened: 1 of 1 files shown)


exelban:
and how that fixes the problem?

arnaldobatista:
On macOS 26.2, SecCodeCopySigningInformation fails with status -67062, so codeSigningMatches(pid:) throws and shouldAcceptNewConnection rejects the XPC. That prevents the app from connecting to the helper, so it can’t call the SMC routines to change the fan. The fix treats -67062 as non‑fatal in the helper, allowing the XPC connection to proceed. With the connection restored, the app can invoke setFanSpeed via the helper and fan control works again.

exelban:
connection with XPC should not be established if code signature do not match.

PR #2917, opened 2026-01-18
Add country flag display for public IP in network widget

jestfortest:


exelban:
Sorry, but there are several issues with this PR:
- You’re using an image instead of a simple emoji to display the flag.
- For some reason you *download* a .png from a third-party service and save it to the system. It’s definitely the worst possible approach for this case. Since it could be done without any downloads. Even in case it could not be done this feature will not be implemented if it requires to download something.

PR #2943, opened 2026-01-31
Update logic CPU frequency detection for M5

dogsteve:
My macbook m5 pro is missing frequency in CPU tab, so i added support for m5 macbook

### DIFF
--- a/Kit/plugins/SystemKit.swift
+++ b/Kit/plugins/SystemKit.swift
@@ -723,9 +723,7 @@ public class SystemKit {
                     return .m4
                 }
             } else if name.contains("m5") {
-                if name.contains("pro") {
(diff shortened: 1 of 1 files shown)


exelban:
please clean the project files

dogsteve:
i haved updated and removed project files

Kit/plugins/SystemKit.swift
```diff
@@ -714,6 +722,10 @@ public class SystemKit {
                 } else {
                     return .m4
                 }
+            } else if name.contains("m5") {
+                if name.contains("pro") {
+                    return .m5
```
exelban: so it's m5 or m5 pro?

PR #2988, opened 2026-02-26
Fix stale popup data when reopening a module popup

### LINKED ISSUE #2140: [Feature Request] Refresh data when clicking menu bar item.
Hello, I don't know whether this is a bug or a feature request (I suppose the latter). I like to keep a "high" refresh interval (5s) to avoid excessive energy usage, but when I click on the menu bar item to view the details in the pop-up, I think it would make sense for the stats to refresh immediately (i.e., to force an update), as we can assume that in this case the user is actively monitoring the system.
Thank you.

eyupcanakman:
Fixes #2140.

## Summary
High update intervals leave popup values stale on reopen. `Reader.start()` only primes a fresh read during initial setup, so when a popup reader pauses on close and is restarted on open, it waits for the next scheduled tick before showing current data. Add a branch that dispatches a background read when a popup or sleep reader reopens while inactive.

## Testing
- [x] `xcodebuild -project Stats.xcodeproj -scheme Stats -destination 'platform=macOS' -configuration Release archive CODE_SIGNING_ALLOWED=NO`
- [x] Verified on macOS: popup refreshes immediately on reopen with a 5s update interval

### DIFF
--- a/Kit/module/reader.swift
+++ b/Kit/module/reader.swift
@@ -148,6 +148,9 @@ open class Reader<T: Codable>: NSObject, ReaderInternal_p {
                 self.repeatTask?.start()
             }
             self.initlizalized = true
+        } else if (self.popup || self.sleep) && !self.active {
+            DispatchQueue.global(qos: .background).async { self.read() }
+            self.repeatTask?.start()
         } else {
             self.repeatTask?.start()
         }


jestfortest (approved):

exelban:
I'm sorry it's taken so long. I never had the time to test that. Thanks for your contribution.

PR #3005, opened 2026-03-05
fix: ensure full hitbox for combined modules when popup is disabled

### LINKED ISSUE #2998: Combined Modules hitbox is smaller
**Describe the bug**
The hitbox of the combined module is slightly smaller when "combined details" is off. (see video)
This is quite frustrating (especially the top hitbox) because I'm used to just move my mouse as far up as possible...

Thank you!

https://github.com/user-attachments/assets/923ac329-1d97-43a9-9f4e-20250975c455

**Details:**
 - Device: Macbook Pro M5
 - macOS: Tahoe 26.3
 - Application version: 2.12.0

Seungwan98:
## Description

Fixes #2998

When combined modules popup is disabled (individual module popups enabled), the hitbox was smaller than expected because individual widgets have margins and don't fill the entire menu bar height. This made the top area unclickable, which was frustrating for users who expect to be able to move their mouse to the top of the screen.

## Changes

- Added a new `handleClick` method that captures clicks across the entire button area
- The method determines which widget was clicked based on mouse position and forwards the click appropriately
- Updated both initial setup and dynamic mode switching to use the new handler

## Testing

1. Enable Combined Modules
2. Disable 'Combined modules popup' in settings (so individual module popups are used)
3. Click on the top area of the combined module in the menu bar
4. The appropriate module popup should now open

## Technical Details

Individual widgets are positioned with a y-offset (`Constants.Widget.margin.y`) and have reduced height, leaving empty space at the top. The button was not capturing clicks in this area when popup mode was disabled. This fix ensures the button always captures clicks and forwards them to the correct widget.

### DIFF
--- a/Stats/Views/CombinedView.swift
+++ b/Stats/Views/CombinedView.swift
@@ -92,6 +92,9 @@ internal class CombinedView: NSObject, NSGestureRecognizerDelegate {
                     }
                 }
             }
+            self.menuBarItem?.button?.target = self
+            self.menuBarItem?.button?.action = #selector(self.handleClick)
+            self.menuBarItem?.button?.sendAction(on: [.leftMouseDown, .rightMouseDown])
         } else {
             self.menuBarItem?.button?.target = self
             self.menuBarItem?.button?.action = #selector(self.togglePopup)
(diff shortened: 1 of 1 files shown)


exelban:
I have the same behavior in with this fix as on #2998 film

PR #3023, opened 2026-03-12
fix: show adapter current and voltage with correct units

bhouston:
Read adapter current/voltage from external power details and only fall back to charger data when needed, so current is populated reliably. Format charging values in the popup as A/V instead of mA/mV to match what users expect.

Before this change on Tahoe the battery details where:

Screenshot 2026-03-12 at 12 01 38 PM

After this change, the battery details are:

Screenshot 2026-03-12 at 11 57 53 AM

Disclaimers:

1. This is my first MacOS / swift contribution and it was AI-assisted.
2. I have only tested on the current version of Tahoe (Version 26.3.1 (25D2128)) on a MacBook Air M3.
3. I didn't see any CONTRIBUTOR.md file so I am making this a bit blindly.

### DIFF
--- a/Modules/Battery/popup.swift
+++ b/Modules/Battery/popup.swift
@@ -302,8 +302,8 @@ internal class Popup: PopupWrapper {
             
             self.powerField?.stringValue = value.isBatteryPowered ? localizedString("Not connected") : "\(value.ACwatts) W"
             self.chargingStateField?.stringValue = value.isCharging ? localizedString("Yes") : localizedString("No")
-            self.chargingCurrentField?.stringValue = value.isBatteryPowered ? localizedString("Not connected") : "\(value.chargingCurrent) mA"
-            self.chargingVoltageField?.stringValue = value.isBatteryPowered ? localizedString("Not connected") : "\(value.chargingVoltage) mV"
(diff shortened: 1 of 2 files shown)


Modules/Battery/popup.swift
```diff
@@ -302,8 +302,8 @@ internal class Popup: PopupWrapper {
             
             self.powerField?.stringValue = value.isBatteryPowered ? localizedString("Not connected") : "\(value.ACwatts) W"
             self.chargingStateField?.stringValue = value.isCharging ? localizedString("Yes") : localizedString("No")
-            self.chargingCurrentField?.stringValue = value.isBatteryPowered ? localizedString("Not connected") : "\(value.chargingCurrent) mA"
-            self.chargingVoltageField?.stringValue = value.isBatteryPowered ? localizedString("Not connected") : "\(value.chargingVoltage) mV"
+            self.chargingCurrentField?.stringValue = value.isBatteryPowered ? localizedString("Not connected") : "\((Double(value.chargingCurrent) / 1000).roundTo(decimalPlaces: 2)) A"
+            self.chargingVoltageField?.stringValue = value.isBatteryPowered ? localizedString("Not connected") : "\((Double(value.chargingVoltage) / 1000).roundTo(decimalPlaces: 2)) V"
```
exelban: only these lines must be changed according to the PR. What changes in reader for?

PR #3024, opened 2026-03-12
feature: Add wattage when adapter attached.

bhouston:
This is build upon my previous PR here that fixed a data collection bug: https://github.com/exelban/stats/pull/3023

I added this option in the battery section:

Screenshot 2026-03-12 at 1 47 01 PM

It then displays as a wattage (only when an adapter is plugged in) in the menu bar like this:

Screenshot 2026-03-12 at 1 44 33 PM

Or like this:

Screenshot 2026-03-12 at 1 43 55 PM

This is important to me so that I know if I am on an under powered adapter - there are so many different wattages of power adapters these days, and sometimes the USB cables you use also affect this.

Disclaimers:

1. This is my second MacOS / swift contribution and it is AI assisted.
2. I have only tested on the current version of Tahoe (Version 26.3.1 (25D2128)) on a MacBook Air M3.
3. I didn't see any CONTRIBUTOR.md file so I am making this a bit blindly.

### DIFF
--- a/Kit/Widgets/Battery.swift
+++ b/Kit/Widgets/Battery.swift
@@ -19,12 +19,14 @@ public class BatteryWidget: WidgetWrapper {
     private var hideAdditionalWhenFull: Bool = true
     private var xlSizeState: Bool = false
     private var chargerIconInside: Bool = true
+    private var adapterWattage: Bool = false
     
     private var _percentage: Double? = nil
     private var _time: Int = 0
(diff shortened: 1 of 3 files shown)


bhouston:
@exelban You didn't like this feature?  I thought it was useful and it fits well into the menu bar at this location.

bhouston:
I could redo it upon your other changes now that the power readings for Apple Silicon are fixed?

PR #3027, opened 2026-03-13
Add base M5 CPU and GPU core sensors

MiikaLaurila:
I was trialing with various SMC scripts I found attached to previous sensor update PRs/issues and took the that as basis to try to deduce the sensors on a base M5 Macbook Air. (10 CPU cores - 8 GPU Cores).

## CPU
I used both custom python scripts and Cinebench to try to differentiate the different cores. For P core tests, I ran a simple busyloop with 4 instances and those seemed to always land on the P cores. For E core tests, I ran same busyloop with 6 instances, and`taskpolicy -c background` to keep all the tasks on the E cores. E cores proved to be quite hard to discern this way, since the power usage of them is so low that the temperatures barely change. I thus also tried to run Cinebencg 2026 all core test with `taskpolicy -c background` and got a bit better results, but it is still quite fuzzy. The power only goes from like 3 -> 4.5 watts with this taskpolicy method.

For "all core" tests I ran the python script with 10 instances without any taskpolicy, and also Cinebench with all core render test. There is a interesting difference in core utilization with the simple Python busyloop test with 10 separate processes and Cinebench with more complex load and within single process. The Cinebench test uses 5-7W more power, but the P cores stay almost 10c cooler, while the E cores get way hotter. I don't know enough about MacOS internals to comment further, but I can only guess there is some optimization either on Apple's or Cinebench's side to optimize benchmark. Or maybe it is just an OS optimization to reduce load on P cores when there is a task that truly fully utilizes the CPU capabilities.

### CPU extras
In addition to the PR's definitions, there are 4 extras in CPU section, that seem to somewhat correlate with the CPU load.

- `Tp0O` seems to somewhat match the calculated average of the performance core values
- `Tp0R` seems to somwwhat match the average of all cores
- `Tp0p` and `Tp0u` are both in the P core range, but seem to have no other immediately obvious correlations

## GPU
These are most likely wrong, but at least a starting point maybe... Method included running `osx-bench` GPU test on loop and Cinebench 2026 GPU render test with multiple trials. These 8 sensors in the PR were the ones that consistently were the 8 hottest ones. Especially in Cinebench the delta between the 8th sensor and 9th sensor was around 1 degrees. In otherwords, in Cinebench a clear pattern of 8 cores could be separated from the Tg sensor set.

#### Graph
full_p_core

#### Utilization
full_p_core_util

#### Graph (cinebench)
full_e_core

#### Utilization (cinebench)
full_e_core_util

#### Graph (python)
full_all_core_python

#### Utilization (python)
full_all_core_python_util

#### Graph (cinebench)
full_all_core_cinebench

#### Utilization (cinebench)
full_all_core_cinebench_util

### DIFF
--- a/Modules/Sensors/values.swift
+++ b/Modules/Sensors/values.swift
@@ -464,7 +464,6 @@ internal let SensorsList: [Sensor] = [
     Sensor(key: "Tm0p", name: "Memory Proximity 1", group: .sensor, type: .temperature, platforms: Platform.m4Gen),
     Sensor(key: "Tm1p", name: "Memory Proximity 2", group: .sensor, type: .temperature, platforms: Platform.m4Gen),
     Sensor(key: "Tm2p", name: "Memory Proximity 3", group: .sensor, type: .temperature, platforms: Platform.m4Gen),
-
     
     // M5
     Sensor(key: "Tp00", name: "CPU efficiency core 1", group: .CPU, type: .temperature, platforms: Platform.m5Gen, average: true),


exelban:
looks very nice, thx, I will review that tomorrow

MiikaLaurila:
Additional info I forgot to write:

I started the search with the Te* sensors, since it would make sense from naming scheme, but when trying to plot them some were in the P core range and some in the E core range. So it might be the case that simply one of the Te values needs to be used for all E cores and there is no way to get a sensor for all cores (meaning the Tp sensors mapped in this PR are just coincidentally correlating or something).

Ta* sensors were a complete dead end. All the Ta sensors report same value in lockstep, and it seems to be an average of some sorts. It matches very well what you get when you average _all_ Tp sensors, but still differently than`Tp0R` that I mentioned in the PR description.

I'd love for someone else to run these trials too since I am getting the tetris effect from all these keys and value spreadsheets I'm collecting :D

exelban:
maybe you can share the scripts you used?
or maybe even make small instruction of how it could be done?

it will be very useful not only for next m gen, but also for pro/max/ultra, since it's very possible that they will have more sensors

MiikaLaurila:
Sure, I can put them to a new repo on Sunday (tomorrow can't). But basically the idea is based on the already existing discovery script found in the M4 thread, that uses the SMC util in Stats. Just extended with heuristics, csv and plot generation etc.

Edit: as for the benchmark script it is just a simple busyloop infinitely doing small math operations

PR #3039, opened 2026-03-17
feat: added drag-and-drop reordering for sensors popup items

omniwired:
- Added drag-and-drop reordering for individual sensors within each type group (Temperature, Voltage, Power, etc.) in the Popup settings tab
- Added drag-and-drop reordering for entire sensor categories (drag the section header grip handle)

Looks like this:

image

### DIFF
--- a/Modules/Sensors/popup.swift
+++ b/Modules/Sensors/popup.swift
@@ -33,12 +33,12 @@ private struct Sensor_t: KeyValue_p {
 
 internal class Popup: PopupWrapper {
     private var list: [String: NSView] = [:]
-    
+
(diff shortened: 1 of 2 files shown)


omniwired:
got it :) 
well if anyone else is interested I will be making a version with this and some new graphs in here
https://github.com/omniwired/stats/tree/master

exelban:
just out of curiosity, why you decide to go with total custom view instead of using the existing approach?

omniwired:
I tried using the one from the clock first, but it felt clunky and unpolished, so I felt this approach looks better easier to grasp.

Followed the same pattern used for widget reordering in the codebase — NSTableView doesn't handle the two-level drag (sections + rows) as cleanly

PR #3051, opened 2026-03-18
fix: fixed fan control helper on M5

c-herz:
Fan control is broken on M5 gen. Setting mode or speed through the UI or smc CLI does nothing, even though RPM reads work fine.

Two issues: M5 uses F%dmd (lowercase) for the fan mode key instead of F%dMd, and the uppercase key doesn't exist so writes are silently ignored. Second, Ftst doesn't exist on M5, but unlockFanControl bails out immediately if it can't read Ftst.

Changes:
- Added fanModeKey() on SMC that probes F0md once and caches the result, returning the correct key name per platform (Intel always gets uppercase at compile time).
- Restructured unlockFanControl to try a direct mode key write first (which is all M5 needs), then fall back to the Ftst unlock mechanism for M1-M4.
- resetFanControl similarly handles missing Ftst by resetting each fan's mode key directly.
- Extended the getValue zero-byte exemption list with the lowercase variants (F0md, F1md).
- Replaced all hardcoded F%dMd references across smc.swift, readers.swift, and main.swift with fanModeKey().

Tested on MacBook Pro 14" M5 Pro (15-core CPU): manual mode, speed control, and reset all work correctly through both the CLI and the Stats UI. Would appreciate confirmation on base M5, M5 Max, and M1-M4 to make sure nothing regressed.

### DIFF
--- a/Modules/Sensors/readers.swift
+++ b/Modules/Sensors/readers.swift
@@ -332,7 +332,7 @@ extension SensorsReader {
                 }
             }
             
-            if let md = SMC.shared.getValue("F\(i)Md"), let parsed = FanMode(rawValue: Int(md)) {
+            if let md = SMC.shared.getValue(SMC.shared.fanModeKey(i)), let parsed = FanMode(rawValue: Int(md)) {
                 mode = parsed.isAutomatic ? .automatic : parsed
             } else {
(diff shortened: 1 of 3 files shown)


exelban:
can confirm that it works fine for m1

PR #3134, opened 2026-04-13
Ram pie chart - Add pressure toggle and Box removal toggle

splady:
Memory pressure mode for the pie chart widget: When enabled, the RAM pie chart displays the system's memory pressure percentage (via memorystatus_get_level() syscall) as a single color-coded segment (green/yellow/red) instead of the app/wired/compressed breakdown
Box toggle for the pie chart widget: Matches the existing BarChart/LineChart pattern — when disabled, the pie chart renders with a transparent background to blend into the menu bar

Here are the screenshots of how these toggles affect the display

Box On, Monochrome, Pressure On Light Black
Box On, Monochrome, Pressure On Light Black  <br>
Box On, Monochrome, Pressure On Light White
Box On, Monochrome, Pressure On Light White  <br>
Box On, No Monochrome, Pressure On, Light White
Box On, No Monochrome, Pressure On, Light White  <br>
No Box, Monochrome Off, Pressure Off
No Box, Monochrome Off, Pressure Off  <br>
No Box, Monochrome, Pressure On Light Black
No Box, Monochrome, Pressure On Light Black  <br>
No Box, Monochrome, Pressure On Light White
No Box, Monochrome, Pressure On Light White


exelban:
hi, thanks for your contribution. 

Few moments:
- 1 & 2 monochrome ok, but 3-5 not ok, it should be grayscale not one value
- pressure should not be as a value but as a colorize option (the same as in the other widgets)
- pressure option should be only for RAM, there is no pressure in other modules
- Box option is incorrect here, for Pie widget its not a box, but you just remove inactive value from the chart. Correct then will be something like -> hide inactive space. The main difference that in the Box does not contain any additional information, your implementation of Box for Pie widget is totally different.
- pressure percentage looks fine but I'm aware about this read across different Macs (intel?)
- and the last thing, you make like 5+ pretty big changes in this PR, and for 100% I will not accept that in this form. It should be split to separate PRs

PR #3160, opened 2026-04-21
fix: improve RAM usage calculation and top processes

agentHits:
image

This PR fixes two RAM-related issues:

- the used memory value is now closer to Activity Monitor
- the top processes list now uses more accurate RAM data and preserves real process names in non-combined mode

### Changes

- adjusted RAM breakdown calculation in `Modules/RAM/readers.swift`
- switched top process memory usage to `proc_pid_rusage(..., ri_phys_footprint)`
- improved process name resolution for helper/web content processes
- refreshed the RAM process list immediately after RAM settings changes

### Tests

- updated `Tests/RAM.swift`
- verified with:

```bash
xcodebuild test \
  -project Stats.xcodeproj \
  -scheme Stats \
  -destination 'platform=macOS' \
  -only-testing:Tests/RAM \
  CODE_SIGNING_ALLOWED=NO \
  CODE_SIGNING_REQUIRED=NO \
  CODE_SIGN_IDENTITY=''

### DIFF
--- a/Modules/RAM/main.swift
+++ b/Modules/RAM/main.swift
@@ -118,6 +118,7 @@ public class RAM: Module {
         
         self.settingsView.callback = { [weak self] in
             self?.usageReader?.read()
+            self?.processReader?.read()
         }
         self.settingsView.setInterval = { [weak self] value in
             self?.processReader?.read()
--- a/Modules/RAM/main.swift
+++ b/Modules/RAM/main.swift
@@ -118,7 +118,6 @@ public class RAM: Module {
         
         self.settingsView.callback = { [weak self] in
             self?.usageReader?.read()
-            self?.processReader?.read()
         }
         self.settingsView.setInterval = { [weak self] value in
             self?.processReader?.read()
--- a/Modules/RAM/readers.swift
+++ b/Modules/RAM/readers.swift
@@ -10,11 +10,7 @@
 //
 
 import Cocoa
-import Darwin
 import Kit
-
-@_silgen_name("proc_pid_rusage")
-private func proc_pid_rusage_bridge(_ pid: Int32, _ flavor: Int32, _ buffer: UnsafeMutableRawPointer) -> Int32
 
 internal struct RAMMemoryBreakdown {
     let app: Double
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: Modules/RAM/main.swift [hunk 1]
```diff
@@ -118,6 +118,7 @@ public class RAM: Module {
         
         self.settingsView.callback = { [weak self] in
             self?.usageReader?.read()
+            self?.processReader?.read()
```
exelban: why open settings trigger process read?

Modules/RAM/readers.swift
```diff
@@ -10,8 +10,27 @@
 //
 
 import Cocoa
+import Darwin
 import Kit
 
+@_silgen_name("proc_pid_rusage")
```
exelban: please avoid using @_silgen_name

exelban:
too many changes in one PR. You not only trying to improve RAM usage calculation, but also rewrite the whole top process logic. Too much in one PR.

agentHits:
> too many changes in one PR. You not only trying to improve RAM usage calculation, but also rewrite the whole top process logic. Too much in one PR.

Understood. I mixed separate concerns in this PR.

The process read from settings was only meant to refresh the top processes immediately after changing process-related settings, but I agree it should not be triggered from opening/settings callback. I removed that.

I also removed the proc_pid_rusage / @_silgen_name based top-process rewrite and the process tree changes from this PR. This PR is now scoped to the RAM usage calculation only.



________
I wrote the entire code using Codex. I don't know Swift. I saw a bug and decided to try to solve it using AI. Yes, that partially solved the problem. Thanks for your comments. I think I've completely fixed it with your answers. Please check.

Modules/RAM/main.swift
agentHits: The intent there was to refresh the top process list immediately after toggling RAM-related settings, especially the combined processes option.

You're right though: `settingsView.callback` is too broad for that and can also be triggered when opening/settings refresh. I'll remove the process read from this callback and keep it scoped only to the setting that actually affects the process list.

Modules/RAM/readers.swift
agentHits: Understood. I'll remove `@_silgen_name` from this PR.

PR #3169, opened 2026-04-26
feat(net): add manual public IP refresh and display proxy exit IP

mattan8844:
- add a manual "Refresh" action for public IP in Net settings
- keep public IP related settings rows in sync with the Public IP toggle
- detect scoped/global HTTP, HTTPS, and SOCKS proxy settings
- fetch and store the external IP seen through the configured proxy
- show a new "Proxy IP" row in the Net popup Address section
- refresh direct and proxied public IP values together
Xnip2026-04-26_18-13-36

### DIFF
--- a/Modules/Net/main.swift
+++ b/Modules/Net/main.swift
@@ -74,6 +74,7 @@ public struct Network_Usage: Codable, RemoteType {
     
     var laddr: Network_addr = Network_addr() // local ip
     var raddr: Network_addr = Network_addr() // remote ip
+    var paddr: Network_addr? = nil // remote ip via configured proxy
     
     var dns: [String] = []
     
@@ -88,6 +89,7 @@ public struct Network_Usage: Codable, RemoteType {
         
         self.laddr = Network_addr()
         self.raddr = Network_addr()
+        self.paddr = nil
         
         self.dns = []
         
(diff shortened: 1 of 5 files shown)


exelban:
what do you mean `add manual public IP refresh`?
this feature exists for a years.

mattan8844:
> what do you mean `add manual public IP refresh`? this feature exists for a years.

image
Add a new button here (as screenshot) so users can refresh the IP immediately by clicking the button.
Also, display the proxy IP when a global HTTP/HTTPS proxy is configured on the system.

exelban:
Zrzut ekranu 2026-04-26 o 12 28 06

exelban:
what are you talking about?

mattan8844:
> what are you talking about?

Sorry, that was my mistake. I canceled it.

exelban:
please explain me an idea of these proxy ip

mattan8844:
> please explain me an idea of these proxy ip

If a global HTTP/HTTPS proxy is configured on the system, the proxy IP will be displayed in the address section.
image

exelban:
I understand that, but I'm not sure I see a reason to have it in the Stats. Like what the case when you need to have that in the Stats?

mattan8844:
> I understand that, but I'm not sure I see a reason to have it in the Stats. Like what the case when you need to have that in the Stats?

For users who do not use a proxy, this feature does not change the current experience. If no global proxy is configured, the row of "proxy IP" will not be shown in the widget.

For users who frequently connect to a proxy and switch proxies from time to time, displaying the proxy IP in the widget makes it easier for them to see which proxy they are currently connected to.

PR #3170, opened 2026-04-26
feat(net): display proxy ip on Net widget

mattan8844:
If a global HTTP/HTTPS proxy is configured on the system, the proxy IP will be displayed in the address section.

For users who frequently connect to a proxy and switch proxies from time to time, displaying the proxy IP in the widget makes it easier for them to see which proxy they are currently connected to.

***For users who do not use a proxy, this feature does not change the current experience. If no global proxy is configured, the row of "proxy IP" will not be shown in the widget.

image


agentHits:
In the implementation, it will be more convenient to see if a proxy is connected and what IP it has, and to see what IP it has without a proxy - it is more convenient to implement this as a 2-in-1

Issue #3091, opened 2026-03-26
Fan control defaults to manual mode after sleep/wake when set to  automatic


Yzaza:
**Device: MacBook Pro M1 Max**

  **Stats version: 2.12.0**

  **Description:**

  When fan control is set to automatic mode and the Mac goes to sleep, waking
  the Mac causes the fan mode to switch to manual/forced unexpectedly.

  Steps to reproduce:
  1. Install the fan helper and set fan control to automatic
  2. Put the Mac to sleep (close lid or Apple menu → Sleep)
  3. Wake the Mac
  4. Open the Stats fan control popup — fan mode is now set to manual

  Expected behavior: Fan mode should remain on automatic after wake.

  Root cause (from code reading):

  In Modules/Sensors/popup.swift, the wakeListener at line 841 unconditionally
  sets resetModeAfterSleep = true, regardless of what mode the fan was in before
   sleep.

  After wake, when the SMC briefly reports a non-automatic mode during hardware
  re-initialization, the update callback at line 918 sees resetModeAfterSleep ==
   true && !value.mode.isAutomatic, and calls modeButtons?.setMode(.forced) —
  which triggers the callback that persists forced mode to UserDefaults and
  writes it to SMC.

  The sleepListener already guards against automatic mode (line 871), so it
  correctly does nothing when the fan is automatic. But the wake listener
  doesn't have the same guard.

  Suggested fix: Move resetModeAfterSleep = true from wakeListener into
  sleepListener (inside the existing guard), so the flag is only set when the
  fan was actually in a non-automatic mode before sleep.

Yzaza:
Closing this — after updating to 2.12.6, the issue is resolved. The fix was
  included in #2983, where == .automatic comparisons were replaced with
  .isAutomatic to properly handle the auto3 (mode 3) fan mode that the SMC
  reports after wake on Apple Silicon. Thanks!

Issue #2977, opened 2026-02-23
fan control spin fan to 100% after sleep


Minhao-Geng:
**Describe the bug**
fans will reach its maximum speed automatically every time I login to the system. In the sensor's popup window,  two fans will switch to Manual options with 100% speed.

**Details:**
 - Device: Macbook Pro with M4Max
 - macOS: 15.7.4
 - Application version: 2.12.0

exelban:
do you have fans in manual or auto mode ?
also do you have enabled to save fans speed?

Minhao-Geng:
> do you have fans in manual or auto mode ? also do you have enabled to save fans speed?

they are in auto mode and i haven't enabled 'save fans speed' yet. 

the problem is every time i logged in, the fans are switched from audo mode to manual mode with 100% speed settings on

Minhao-Geng:
> do you have fans in manual or auto mode ? also do you have enabled to save fans speed?

i tried enabling the "save fans speed" but it doesn't work.

you can try sleeping your mac first and log in to wait for a few seconds, you will hear the noise from fans. at the same time if you check the popup window, you probably notice the fans are on manual mode with 100% speed

exelban:
nope, I don't have this issue

exelban:
maybe @engelberger could help?

Minhao-Geng:
> maybe [@engelberger](https://github.com/engelberger) could help?

i reinstalled the software but still doesn't help. 

the issue is slightly different this time, where fans spin to minimum it can be instead of maximum. 

anyway, the main issue is fans will enable manual mode automatically after sleep

exelban:
Could you try uninstall and install again the fan helper?

Minhao-Geng:
> Could you try uninstall and install again the fan helper?

i already tried everything like uninstall fan helper, uninstall the entire software using app cleaner and restart to reinstall but noting works.

the issue sometimes cannot be reproduced but it is highly likely to happen if i click the sensor's popup immediately after the login

Will-Georges:
I am having the exact same thing. I originally thought there was a problem with my mac because fan control hasn't worked in stats for ages but seems to have been re added. Seems to only have started happening on 2.12.0

engelberger:
hey, i looked into this and i think i found the bug. it seems like a regression from PR #2924 (the Apple Silicon fan control changes in 2.12.0).

[...]

PR #3189, opened 2026-05-07
fix(fans): use IOKit power notifications for sleep/wake fan state

marxo126:
## Problem

Fan sleep/wake handling currently lives in `Modules/Sensors/popup.swift` and uses `NSWorkspace.didWakeNotification` / `NSWorkspace.willSleepNotification`. This has three structural issues:

1. `NSWorkspace.didWakeNotification` arrives after SMC is ready only intermittently — `IORegisterForSystemPower` (IOKit) fires earlier and more reliably.
2. The sleep/wake observers are added on the popup view (`FanView`). If the user closes the menu bar popup between sleep and wake, the observers are torn down with the view and the user's manual fan mode is never restored.
3. `willSleepMode` / `willSleepSpeed` snapshot lives on the popup view, so it is only valid while the popup is alive.

This causes the regressions reported in #3091 ("manual mode resets to auto after sleep/wake") and #2977 ("fan goes to 100% after sleep" — same root cause path: snapshot lost, restoration falls back to default).

## Change

- New `Modules/Sensors/fanPower.swift` — `FanPowerManager.shared` singleton subscribing to `IORegisterForSystemPower`. Holds the per-fan `(mode, speed)` snapshot dictionary at the app level, independent of any view lifecycle.
- On `kIOMessageSystemWillSleep`: snapshot every registered fan with non-automatic mode, then set automatic before sleep. Acknowledge with `IOAllowPowerChange`.
- On `kIOMessageSystemHasPoweredOn`: 2-second delay (let SMC settle), then restore mode + speed via `SMCHelper.shared`. Clear snapshot.
- `Modules/Sensors/popup.swift` — removed the `NSWorkspace` observers, `wakeListener`, `sleepListener`, `willSleepMode`, `willSleepSpeed`, and the now-dead `resetModeAfterSleep` re-sync block. The reader's normal tick picks up the restored mode on next read.
- `Modules/Sensors/main.swift` — registers each discovered fan with `FanPowerManager.shared` at module init time.

## Net diff

```
4 files changed, 121 insertions(+), 55 deletions(-)
 Modules/Sensors/fanPower.swift  | 114 ++++++++++++++++ (new)
 Modules/Sensors/main.swift      |   3 +
 Modules/Sensors/popup.swift     | -55
 Stats.xcodeproj/project.pbxproj |   4 +
```

## Testing

Logic-reviewed against the existing fan code paths but **not yet tested on real hardware** — flagging this as Draft so you can take a look at the approach before I run it through sleep/wake on Apple Silicon.

If the approach lands well I'm happy to:
- Run a sleep/wake matrix on Apple Silicon (M-series) and report
- Add an integration shim if you want this gated behind a settings toggle for safety during rollout

Marking Draft. Happy to iterate or split further if you'd prefer the new singleton lives elsewhere (e.g. `Kit/`).

### DIFF
--- a/Modules/Sensors/fanPower.swift
+++ b/Modules/Sensors/fanPower.swift
@@ -1,0 +1,114 @@
+//
+//  fanPower.swift
+//  Sensors
+//
+//  Copyright © 2024 Serhiy Mytrovtsiy. All rights reserved.
+//
+
+import IOKit
+import IOKit.pwr_mgt
+import Kit
+
+// Manages fan mode/speed snapshot across sleep/wake cycles at the IOKit layer.
+// NSWorkspace.didWakeNotification arrives after SMC is ready only intermittently;
+// IORegisterForSystemPower fires earlier and is not tied to any UI lifecycle.
+internal final class FanPowerManager {
+    internal static let shared = FanPowerManager()
+
+    // Snapshot keyed by fan id: (mode, speed) captured just before sleep.
+    private var snapshot: [Int: (mode: FanMode, speed: Int)] = [:]
+
+    // Fan list registered by the reader on discovery.
+    private var fans: [Fan] = []
+
+    private var notifyPort: IONotificationPortRef?
+    private var notifierObject: io_object_t = 0
+    private var rootPort: io_connect_t = IO_OBJECT_NULL
+
+    private init() {
+        self.registerIOPower()
+    }
+
+    internal func register(fan: Fan) {
+        guard !self.fans.contains(where: { $0.id == fan.id }) else { return }
+        self.fans.append(fan)
+    }
+
+    private func registerIOPower() {
+        // Retain self for the C callback context; released in deinit via IODeregisterForSystemPower.
+        let selfPtr = Unmanaged.passRetained(self).toOpaque()
+
+        let callback: IOPowerSourceCallbackType = { refcon, _, messageType, messageArgument in
+            guard let refcon else { return }
+            let manager = Unmanaged<FanPowerManager>.fromOpaque(refcon).takeUnretainedValue()
+            switch Int(messageType) {
+            case kIOMessageSystemWillSleep:
+                manager.handleWillSleep()
+                IOAllowPowerChange(manager.rootPort, Int(bitPattern: messageArgument))
+            case kIOMessageSystemHasPoweredOn:
+                manager.handleHasPoweredOn()
+            default:
+                break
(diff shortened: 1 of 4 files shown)


marxo126:
**Status update — runtime-verified at the init layer; sleep/wake event not exercised in CI.**

### What was tested on macOS 26.4.1, M4 Max

- App launches with no crash, `FanPowerManager.shared` initialises, `IORegisterForSystemPower` returns a valid `rootPort`, the dispatch queue is wired
- `register(fan:)` callbacks fire from `Modules/Sensors/main.swift` at module init for both real fans (id 0, id 1)
- `popup.swift` no longer references the removed `wakeListener` / `sleepListener` / `willSleepMode` / `willSleepSpeed` / `resetModeAfterSleep` ivars — clean compile, zero deprecation warnings

### What was NOT tested

- The sleep → wake → restore round-trip itself. Putting the test machine to sleep mid-session was disruptive. The actual `kIOMessageSystemHasPoweredOn` callback path is logic-reviewed but not exercised on hardware in this PR. The 2-second post-wake delay before SMC restore is a literal copy of the pattern that fixed similar issues elsewhere; happy to add an `os_log` instrumented verification run if you want a recorded trace.

### Maintainability notes

- The state (per-fan `(mode, speed)` snapshot) lives on the singleton, not on any view. If you ever rip out the popup again or change `FanView`'s lifecycle, sleep/wake handling stays intact.
- `FanPowerManager` deinitialises via `IODeregisterForSystemPower` + `IONotificationPortDestroy`. App quit path is clean.
- The IOKit notification fires on a background queue (`DispatchQueue.global(qos: .utility)`); SMC writes within the handler use `SMCHelper.shared` which is already thread-safe via XPC.
- One narrow assumption: fans returned by the sensor reader's `list.sensors` are stable across the app lifetime. If a fan ever hot-unplugs (rare on Apple Silicon, but possible on external display dock combos) the state dictionary would have a stale entry. Not currently a concern, but a documented unknown.

### Caveats

- Tested with an unsigned dev build (Apple Developer Program account is pending Apple's approval — submitted, awaiting). When the cert lands I will rebuild signed and re-run the sleep/wake matrix; will post results here.
- The fix doesn't depend on signing — `IORegisterForSystemPower` is unrestricted. Runtime behaviour in a Developer-ID-signed production build is identical.

Closes #3091. Closes #2977.

marxo126:

Issue #1166, opened 2022-11-13
Unable to set fan speed


Moulick:
**Describe the bug**
On setting the fan speed in the CPU widget, Stats asks for admin permissions 6 times. It's still is not able to change the fan speed

**Details:**
 - Device: Macbook Pro 14 Inch 2021 M1
 - macOS: Ventura 13.1
 - Application version: 2.7.42

exelban:
Hi. There is no option to change the fan speed throw the CPU module. Maybe you mean the sensors module?
There is no public release of macOS 13.1. So I have no idea why it does not work.

tarushi-01:
> Hi. There is no option to change the fan speed throw the CPU module. Maybe you mean the sensors module? There is no public release of macOS 13.1. So I have no idea why it does not work.

Ah yeah, I mean't the sensor module. I'll re-verify once 13.1 is generally available.

Moulick:
> Hi. There is no option to change the fan speed throw the CPU module. Maybe you mean the sensors module? There is no public release of macOS 13.1. So I have no idea why it does not work.

Ah yeah, I meant the sensor module. I'll re-verify once 13.1 is generally available.

PR #3190, opened 2026-05-07
fix(fans): surface SMC write failures to user instead of silent fail

marxo126:
## Problem

`SMC/smc.swift` exposes `setFanSpeed`, `setFanMode`, and `resetFans` as `Void`. On any failure path — IOKit error, missing key on a chip generation, firmware reject — the function `print()`s and returns silently. The caller in `Modules/Sensors/popup.swift` has no way to know the SMC write failed, so the slider/mode UI stays in the wrong state and the user gets no feedback.

This is the visible symptom in #1166 ("Unable to set fan speed") and a contributing factor to several other fan reports where users report "fan control didn't take effect".

## Change

- `SMC/smc.swift` — `setFanSpeed`, `setFanMode`, `resetFans` now return `@discardableResult Bool`. Every error path returns `false`; success returns `true`. `resetFanControl` already returns `Bool` and is unchanged.
- `SMC/Helper/main.swift` — the existing `(String?) -> Void` completion now receives a non-nil error string on failure (was `nil` for both success and failure, which made errors indistinguishable). Error format: `"Failed to set fan <id> speed to <rpm> RPM — SMC rejected write"`.
- `Kit/helpers.swift` — `SMCHelper` setters gain an optional `((String?) -> Void)?` completion (default `nil` preserves all existing call sites). Forwards XPC error string through.
- `Modules/Sensors/popup.swift` — added `showFanError(_:)` private helper on `FanView` that dispatches an `NSAlert` ("Fan control failed", `.warning` style) to the main queue. All 8 interactive call sites — init restore, mode buttons (auto/manual/off/turbo), `setSpeed` debouncer, `wakeListener` — pass a completion that calls `showFanError` on a non-nil error. `sleepListener` intentionally stays fire-and-forget (the user can't act on a sleep-time error).

## Public API note

Returning `Bool` from previously-`Void` functions is a public API change inside the SMC layer. Marked `@discardableResult` to avoid noise at existing call sites that don't care about the result. If you'd prefer a non-API-changing approach (separate `tryFanSpeed` etc), happy to refactor.

## Net diff

```
4 files changed, +39 lines net
```

## Testing

Logic-reviewed but **not yet hardware-tested**. Marking Draft. The XPC completion fires on an internal XPC queue per Apple's docs, so the alert dispatches via `DispatchQueue.main.async` before `NSAlert.runModal()` — please double-check that pattern matches your house style for the popup layer.

### DIFF
--- a/Kit/helpers.swift
+++ b/Kit/helpers.swift
@@ -913,27 +913,36 @@ public class SMCHelper {
     private var connection: NSXPCConnection? = nil
     
     public func setFanSpeed(_ id: Int, speed: Int, completion: ((String?) -> Void)? = nil) {
-        guard let helper = self.helper(nil) else {
-            completion?("Fan helper not available")
-            return
-        }
+        guard let helper = self.helper(forCall: completion) else { return }
         helper.setFanSpeed(id: id, value: speed) { completion?($0) }
     }
 
     public func setFanMode(_ id: Int, mode: Int, completion: ((String?) -> Void)? = nil) {
-        guard let helper = self.helper(nil) else {
-            completion?("Fan helper not available")
-            return
-        }
+        guard let helper = self.helper(forCall: completion) else { return }
         helper.setFanMode(id: id, mode: mode) { completion?($0) }
     }
 
     public func resetFanControl(completion: ((String?) -> Void)? = nil) {
-        guard let helper = self.helper(nil) else {
-            completion?("Fan helper not available")
-            return
-        }
+        guard let helper = self.helper(forCall: completion) else { return }
         helper.resetFanControl { completion?($0) }
+    }
+
+    // Per-call helper that wires the XPC errorHandler to the caller's completion
+    // so connection failures surface as errors instead of dropping the reply.
+    private func helper(forCall completion: ((String?) -> Void)?) -> HelperProtocol? {
+        guard let connection = self.helperConnection() else {
+            completion?("Fan helper not available")
+            return nil
+        }
+        let proxy = connection.remoteObjectProxyWithErrorHandler { error in
+            completion?("Fan helper unreachable: \(error.localizedDescription)")
+        }
+        guard let service = proxy as? HelperProtocol else {
+            completion?("Fan helper proxy unavailable")
+            return nil
+        }
+        service.setSMCPath(Bundle.main.path(forResource: "smc", ofType: nil)!)
+        return service
     }
     
     public func isActive() -> Bool {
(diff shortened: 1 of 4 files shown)


marxo126:
**Update — runtime verified end-to-end on macOS 26.4.1, Apple Silicon (M4 Max).**

Test setup:
1. \`launchctl bootout system /Library/LaunchDaemons/eu.exelban.Stats.SMC.Helper.plist\` — torn down the helper to force XPC connection failure on the next fan write.
2. Launched dev build, opened popup, clicked **Manual** on a fan.

Result: NSAlert pops with the expected "Fan control failed" + propagated error string ("Fan helper unreachable: Couldn't communicate with a helper application"). Confirmed via popup screenshot.

One real bug surfaced during testing — the original wiring passed the completion only through the proxy's reply block, so XPC connection failures (helper bootstrapped-off, helper crashed) silently dropped the completion and the alert never fired. Fix is in commit \`2bb88f02\`: per-call helper now sets \`remoteObjectProxyWithErrorHandler\` to invoke the caller's completion with a typed error string, so XPC failures take the same surfacing path as helper-side rejects. The fix is what makes this PR actually deliver its promised behaviour.

Caveat unchanged: tested on a (necessarily) unsigned dev build with a matching unsigned helper to bypass the SMAuthorizedClients certificate check. Behaviour in a Developer ID-signed production build is identical — the signature check is independent of the failure-surfacing path.

marxo126:
**Status — additional maintainability notes + Developer Program update.**

PR is functionally verified per the [earlier comment](https://github.com/exelban/stats/pull/3190#issuecomment-4401189597) (NSAlert pops with the propagated XPC error string). Adding context for future maintenance:

### Maintainability notes

- The `Bool` return on `setFanSpeed` / `setFanMode` / `resetFans` is marked `@discardableResult` so existing void call sites (e.g. `wakeListener` in #3189) don't generate warnings.
- The XPC error path goes through `remoteObjectProxyWithErrorHandler` in `Kit/helpers.swift`. Callers that don't pass a completion still log via the helper's existing `print` path; no behavioural change for those.
- `NSAlert` runs on the main queue via explicit `DispatchQueue.main.async` — avoids the XPC reply queue ambiguity flagged by Apple's `NSXPCConnection` docs.
- The `sleepListener` path intentionally does not surface errors to the user (commented in the diff); it would be visible only after wake when there's no actionable correction.

### Apple Developer Program

Account pending Apple approval; will rebuild signed and re-run the bootout-and-click test on a properly notarised binary when the cert arrives. The XPC reply path doesn't differ between signed and unsigned beyond the helper's `SMAuthorizedClients` filter, so signed-build behaviour should match the dev-build verification posted above.

Closes #1166.

marxo126:

PR #3196, opened 2026-05-09
Add thermal pressure state and CPU speed limit to Apple Silicon CPU popup

NocturnalGrp:
Stats covers a lot of ground that other free monitors don't — the Sensors module depth, per-core frequency breakdowns, unified memory breakdown, and GPU and power figures all in one app made it the clear choice for Apple Silicon monitoring. Thanks for maintaining it. This PR builds on that with a small addition for M-series users.

---

### What this adds

On Apple Silicon, `pmset -g therm` is already called by the existing `LimitReader` — but that reader is wrapped in `#if arch(x86_64)`, so on M-series chips the speed limit data is read and discarded. Separately, `ProcessInfo.thermalState` (nominal/fair/serious/critical pressure state) isn't called anywhere in the CPU module.

This PR surfaces both on ARM, adding two new rows to the CPU popup's Details section:

| Row | Source |
|---|---|
| **Thermal pressure** — `Nominal` / `Fair` / `Serious` / `Critical` with a colour-coded dot | `ProcessInfo.processInfo.thermalState`, polled every 5s |
| **Speed limit** — drops below 100% when the chip is thermally throttled | `/usr/bin/pmset -g therm`, same tool already used on Intel |

Intel behaviour is completely unchanged.

### What it looks like

\`\`\`
● Thermal pressure:   Fair
  Speed limit:        95%
\`\`\`

The dot shifts green → yellow → orange → red as pressure increases.

### Implementation

Three files, ~112 lines added:

- **`readers.swift`** — new `ThermalStateReader` class, polls both sources every 5 seconds into a `CPU_ThermalState` struct
- **`main.swift`** — struct definition, reader instantiation in the `#else` arm alongside `FrequencyReader`, callback routing to popup
- **`popup.swift`** — two new rows in `initDetails()` for ARM, `thermalStateCallback` with colour logic, `detailsHeight` unified to 6 base rows for both architectures

No private APIs. No new dependencies.

### Tested on

MacBook Pro M4 Max, 48GB, macOS 26.

### DIFF
--- a/Modules/CPU/main.swift
+++ b/Modules/CPU/main.swift
@@ -44,6 +44,13 @@ public struct CPU_Limit: Codable {
     var speed: Int = 0
 }
 
+public struct CPU_ThermalState: Codable {
+    var thermalPressure: Int = ProcessInfo.ThermalState.nominal.rawValue
+    var schedulerLimit: Int = 100
+    var cpus: Int = 0
+    var speedLimit: Int = 100
+}
+
 public struct CPU_AverageLoad: Codable, RemoteType {
     var load1: Double = 0
     var load5: Double = 0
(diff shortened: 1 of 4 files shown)


psyskeptic:
Any reason why this was not merged? 
Lack of showing cpu throttling is the only function i miss in this app. On Macbook airs laptops which has passive cooling throttling happens a lot

PR #3198, opened 2026-05-09
feat(sensors): temperature-driven fan controller with per-fan AC/Battery profiles

Morteza-Rastgoo:
## What this adds

An opt-in **Fan Temperature Controller** built into the Sensors popup. Instead of manually setting RPM, you pick a **target exhaust temperature** and the controller keeps fans as fast as needed to stay cool — with **separate targets for AC Adapter and Battery**, configurable per fan.

When disabled the engine is completely inert; existing manual/automatic fan behaviour is unchanged.

## Key differences from #3191

| | #3191 (curve engine) | This PR |
|---|---|---|
| UX | Define multi-point RPM curve | Single target temp slider |
| Power source | One profile for all conditions | Separate AC / Battery targets |
| Control sensor | CPU temperature | Exhaust airflow (lap-facing vents) |
| Algorithm | Linear interpolation on user curve | Bang-bang: instant max → slow step-down |

## How it works

Each fan in the Sensors popup gains a third mode button: **Temp**. Clicking it reveals two sliders (⚡ AC and 🔋 Battery), each controlling a target exhaust temperature from 30–85 °C.

**Control signal — exhaust airflow vents (not CPU core temp):**
- Primary: `TaLW` / `TaRW` — left/right wing vent sensors (~32–34 °C at idle on M4 Max)
- Fallback: `TaLP` / `TaRF` then max CPU core temp
- Popup header shows live **L: 32°C  R: 33°C** readout (turns orange when either exceeds target)
- Using vent temp means the controller responds to what actually heats the lap, not die temps that run 60–110 °C above perceived surface heat

**Algorithm:**
- `controlTemp > target` → instantly set **max RPM** (forced mode via SMC)
- `controlTemp ≤ target` → step down 200 RPM every 500 ms until automatic mode is restored
- 500 ms tick gate + `newRPM != prev` guard prevent flooding the helper's serial `smcQueue`
- Mode transitions (Auto/Manual) are handled exclusively by popup buttons; controller does NOT call `setFanMode` for fans not in Temp mode — eliminates a race that would silently override manual speed
- `willTerminate` restores automatic mode on all controlled fans

## Files changed

| File | What changed |
|---|---|
| `Modules/Sensors/fanTempController.swift` | New singleton: power-source detection, airflow sensor selection, bang-bang controller, smcQueue flood prevention (~172 LOC) |
| `Modules/Sensors/popup.swift` | Auto/Manual/Temp mode buttons per fan; Temp panel with ⚡/🔋 sliders; live L/R airflow readout in header (~+207 LOC) |
| `Modules/Sensors/main.swift` | Wire `processTick` / `releaseAll` into Sensors module lifecycle (+8 LOC) |
| `Modules/Sensors/values.swift` | Add `TaLW` / `TaRW` (Airflow left/right wing) to Apple Silicon sensor list (+2 lines) |
| `Stats.xcodeproj/project.pbxproj` | 4 new build entries for `fanTempController.swift` |

## Testing

Verified on macOS 26.4.1, Apple Silicon M2:

- **Temp mode, target 40 °C, laptop on lap**: vent sensors ~32–34 °C at rest → fans stay in automatic; under load (vent temps ~45 °C) → fans blast to 6800 RPM immediately; load drops → step-down over ~20 s back to automatic
- **AC vs Battery profiles**: switching power source instantly applies the matching slider target
- **Manual mode after Temp mode**: clicking Manual exits Temp mode cleanly; slider controls fan speed without controller interference
- **Auto mode**: always restored correctly on quit and on explicit Auto click
- **Disabled (Auto or Manual)**: zero SMC calls from controller; behaviour identical to unmodified Stats


Morteza-Rastgoo:
🙂

PR #3207, opened 2026-05-14
fix(de): correct Gemma 4b mistranslations in Localizable.strings

michi-onl:
Fix 9 incorrect German translations introduced by automated Gemma 4b translation in `Stats/Supporting Files/de.lproj/Localizable.strings`.

### Changes

| Key | Before | After | Reason |
|-----|--------|-------|--------|
| `Number of s-cores` | `%0 Super-Prozessoren` | `%0 S-Kerne` | Pattern consistency with e-cores/p-cores |
| `Box` | `Karton` | `Box` | Established UI loanword |
| `Tachometer widget` | `Drehzahlmesser` | `Tachometer` | Avoids automotive jargon |
| `Load per core` | `Laden pro Kern` | `Last pro Kern` | Matches existing "Last" usage in file |
| `Efficiency core` | `Kern der Effizienz` | `Effizienzkern` | Matches existing "Effizienzkerne" |
| `Performance core` | `Leistungszentrum` | `Leistungskern` | Matches existing "Leistungskerne" |
| `Super core` | `Hervorragende Kernfunktion` | `Superkern` | Semantic error |
| `Upload speed` | `Hochladen` | `Uploadgeschwindigkeit` | "speed" was missing entirely |
| `Download speed` | `Herunterladen` | `Downloadgeschwindigkeit` | "speed" was missing entirely |

### DIFF
--- a/Stats/Supporting Files/de.lproj/Localizable.strings
+++ b/Stats/Supporting Files/de.lproj/Localizable.strings
@@ -178,7 +178,7 @@
 "Number of threads" = "%0 Threads";
 "Number of e-cores" = "%0 Effizienzkerne";
 "Number of p-cores" = "%0 Leistungskerne";
-"Number of s-cores" = "%0 Super-Prozessoren"; // translategemma:4b
+"Number of s-cores" = "%0 S-Kerne";
 "Disks" = "Festplatten";
 "Display" = "Bildschirm";
 
@@ -191,7 +191,7 @@
 // Widgets
 "Color" = "Farbe";
 "Label" = "Beschriftung";
-"Box" = "Karton"; // translategemma:4b
+"Box" = "Box";
 "Frame" = "Rahmen";
 "Value" = "Wert";
 "Colorize" = "Einfärben";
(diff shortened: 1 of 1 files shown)


michi-onl:
Note to the maintainer: Using a SKILL.md file together with a 30B+ model would lead to significantly better results.

exelban:
I would like to run bigger model but have not enough vram for that)

michi-onl:
How were these Gemma translations accomplished? I believe a combination of a good cli harness, full file context, and a skill.md file would be necessary. If these resources are not available, we would need to wait until the models improve or the workflow becomes more accessible.

exelban:
basically I just wanted to fix the problem that a lot of languages has no translations for new keys at all. So even not best translation will be better than no translation of some words at all. I don't want to spend a lot of times for that, but if you want you can help improve that. The translation workflow could be found [here](https://github.com/exelban/stats/blob/75fa7a7a8e2076bee9cc8c60d206aa788c8694c0/Kit/scripts/i18n.py#L180).

michi-onl:
Existing set-and-forget solutions aren't compatible with swift project currently. Despite optimizing the script's context and prompts, local models under 8B parameters failed to follow instructions or output valid structured data with good context awareness. Reliable translations imho currently require a frontier AI model. If cost is a concern, platforms like OpenCode offer limited free access to capable models.

I'll look into improving other locales.

Issue #3212, opened 2026-05-16
[Bug] Invalid Window spam in WindowServer


Incipiens:
Hi there,

First of all, thanks for Stats. I've been using it for a long time, and I covered it at XDA a while ago as it's one of the best quality of life applications that I've used on Mac. 

I'm posting this as a follow-up to #2829. After testing locally on macOS Tahoe 26.4 on a MacBook Pro (M4 Pro), I'm fairly confident Stats is a reliable trigger for the issue, even if the root cause is likely a Tahoe/AppKit/WindowServer regression rather than a simple standalone bug in Stats itself.

While Stats is running:

- WindowServer repeatedly logs:
  - `_CGXPackagesSetWindowConstraints: Invalid window`
- The log rate was roughly 2 to 4 times per second
- WindowServer CPU usage stayed elevated while this was happening
- WindowServer memory usage also appeared to grow over time
- Quitting Stats caused the spam to stop
- I reproduced the same behavior with a local debug build from source

I tried a few different things to try and fix it:

- experimenting with popup/window behavior (`NSWindow` vs `NSPanel`)
- checking popup show/hide paths
- adding guards/instrumentation around widget width/redraw/status-item host validity

However, disabling the normal continuous background update path for most modules and only refreshing them on demand when I open them has fixed the issue.

In my local build, I:
- prevented the normal repeating reader/update loop from starting for most non-battery modules
- manually triggered `read()` for that module's readers when its popup/menu item was opened

With that change, the WindowServer spam was basically stopped in my testing. It still happens rarely, but this is unlikely to be attributed to Stats.

This makes me think the problematic path is more likely the steady-state menu bar update pipeline. In other words, Tahoe seems unhappy with some combination of:

- frequent `NSStatusItem` updates
- dynamic width changes via `item.length`
- custom subviews attached to `statusItem.button`
- repeated relayout/repositioning of menu bar widgets

Stats may just be a particularly strong reproducer because it updates many custom menu bar items frequently.

Hopefully that helps narrow it down a bit. I realize Apple sometimes changes internal behavior here without much warning, so this may be more of a Tahoe compatibility problem that Stats happens to expose reliably.

exelban:
just to clarify cause I'm not sure if I understand correctly, your complain is about Stats updates the widgets in the menu bar?

Incipiens:
> just to clarify cause I'm not sure if I understand correctly, your complain is about Stats updates the widgets in the menu bar?

[...]

Issue #2829, opened 2025-11-22
Tahoe "Invalid window" errors in Console.


dmmalam:
Repeated "Invalid window" errors in Console (2-3 per second) , with stats running. Tried turning widgets on/off one by one to isolate to. Still get some with everything off, then more with each widget correlated to their update interval. Killings stats make them go away.

```
error	18:26:19.182656+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:19.182673+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:19.182708+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:19.721954+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:19.724463+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:19.724580+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:19.811489+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:19.813565+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:19.813585+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.016841+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.025777+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.025809+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.728625+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.730366+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.730436+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.811285+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.813299+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:20.813337+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
error	18:26:21.016787+0000	WindowServer	_CGXPackagesSetWindowConstraints: Invalid window
[...]

PR #3214, opened 2026-05-16
fix: guard menu-bar widget paths against unhosted state

Incipiens:
Several widget redraw and resize paths could fire while a widget was not yet attached to its hosting NSStatusItem button, or while the button itself had no associated NSWindow (windowNumber == -1).

When that happened, calls into setFrameSize/display/NSStatusItem.length would still reach AppKit, which forwarded them to WindowServer and caused log spamming to occur. The same paths were also unconditionally re-adding the widget view to its host on every status-item setup, causing needless view-hierarchy churn on toggle.

This change introduces a small hasUsableHost helper on WidgetWrapper that requires both a non-nil superview and a window with a valid windowNumber, and uses it to early-return from:

- WidgetWrapper.setWidth (before setFrameSize/widthHandler)

- WidgetWrapper.redraw (before display)

For the per-widget status item, SWidget.widthHandler and MenuBar.recalculateWidth now require the menu-bar button to have a window before assigning NSStatusItem.length, which is what actually triggers the WindowServer round-trip and log spam.

SWidget.setMenuBarItem and MenuBarView.addWidget are also made idempotent: they only call addSubview when the view isn't already parented to the intended host, removing repeated reparent work on toggle and one-view changes.

Two unrelated fixes are included in the same change because they surfaced while reproducing the menu-bar issue on macOS Tahoe:

  - Modules/Net/preview.swift: the network preview chart was being initialised without an explicit frame, leaving it sized 0x0 until the first layout pass. Provide an explicit (0, 0, 0, 140) frame so the preview renders correctly on first display.

  - Modules/Net/readers.swift: guard the CWPHYMode.mode11be case in the CustomStringConvertible extension behind `#if compiler(<6.2)`. The symbol is not exposed by the CoreWLAN headers shipped with the Tahoe SDK (Xcode 26/Swift 6.2), so unconditionally referencing it breaks the build on that toolchain. Older toolchains (Sonoma, Sequoia) continue to render Wi-Fi 7 networks as "802.11be".

Related: #2829, #3212

### DIFF
--- a/Kit/module/widget.swift
+++ b/Kit/module/widget.swift
@@ -162,6 +162,12 @@ open class WidgetWrapper: NSView, widget_p {
     public var onClick: (() -> Void)? = nil
     public var shadowSize: CGSize
     internal var queue: DispatchQueue
+
+    private var hasUsableHost: Bool {
+        guard self.superview != nil else { return false }
+        guard let window = self.window else { return false }
+        return window.windowNumber != -1
+    }
     
     public init(_ type: widget_t, title: String, frame: NSRect) {
         self.type = type
@@ -186,6 +192,7 @@ open class WidgetWrapper: NSView, widget_p {
         self.shadowSize.width = newWidth
         
         DispatchQueue.main.async {
+            guard self.hasUsableHost else { return }
             self.setFrameSize(NSSize(width: newWidth, height: self.frame.size.height))
             self.widthHandler?()
         }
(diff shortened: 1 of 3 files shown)


Incipiens:
Actually, closing this. The problem disappeared at launch but then returned two hours later. It's still an improvement, but not the root cause. Apologies for the PR.

PR #3216, opened 2026-05-17
fix: Menu-bar widget guards, bypass setNeedsDisplayInRect

Incipiens:
Firstly, sorry for the earlier PR that was opened. I had thought it addressed the issue, but it actually hadn't. At least not fully. Using lldb, I was able to verify the actual cause.

Breaking on `-[NSStatusItem _windowNeedsReplicantUpdate:]` produced a backtrace that hits `[NSView setNeedsDisplayInRect:]+284` in AppKit, right before the Stats widget code. This was consistent on every Invalid window log entry.

- A widget view (subview of `NSStatusBarButton`) is invalidated by `setNeedsDisplay = true` or `setNeedsDisplayInRect:`.
- `-[NSView setNeedsDisplayInRect:]`, on Tahoe, posts a CoreFoundation notification at offset +284.
- `NSStatusBarWindow` is registered as an observer of that notification (selector `_viewHierarchyDidUpdate`).
- The observer calls `-[NSStatusItem _windowNeedsReplicantUpdate:]`, which pushes a window-geometry constraint to SkyLight via `_CGXPackagesSetWindowConstraints` to update the menu bar's replicant copy of the status item content.
- If the status item's backing window is in any transitional state at that moment (`windowNumber <= 0`, mid-recycle, behind notch, in overflow under Control Center, etc), SkyLight logs `Invalid window` and drops the request.

The trigger is, therefore, any view invalidation inside a status item button. On pre-Tahoe macOS, either the notification was not posted from `setNeedsDisplayInRect:` or `NSStatusBarWindow` did not observe it; on Tahoe it does both, so every widget redraw round-trips through SkyLight.

This PR, as a result, moves widget rendering to layer-backed views with manual `layer.contents`. It renders to a `CGImage` on the main thread and assigns `view.layer.contents = cgImage` directly. This avoids `setNeedsDisplayInRect:` and the notification chain entirely. This reuses each widget's existing draw(_:) by handing it a `CGContext`, so the per-widget drawing code didn't have to change, only the call sites that previously invalidated views. I spent a few hours reading into this yesterday, and it seems as if this is the only mitigation that removes the spam even when items remain visible. 

Transitional states for the status item's backing window are much more frequent on Tahoe than older macOS due to menu bar overflow, the notch, and Liquid Glass animation, so the SkyLight call almost always lands while the window is in one. This fix never calls setNeedsDisplayInRect: in the first place, so the notification never fires and the SkyLight call never happens regardless of what state the window is in.

I measured the impact of this fix on macOS 26.5 over a 10h run: pre-fix ~3.9 hits/sec, post-fix 0.025/sec active and 0.006/sec idle. All APIs used in this fix are available on macOS 11 and newer.

### DIFF
--- a/Kit/Widgets/BarChart.swift
+++ b/Kit/Widgets/BarChart.swift
@@ -83,7 +83,7 @@ public class BarChart: WidgetWrapper {
             }
             self.setFrameSize(NSSize(width: 36, height: self.frame.size.height))
             self.invalidateIntrinsicContentSize()
-            self.display()
+            self.renderToLayer()
         }
         
         let style = NSMutableParagraphStyle()
--- a/Kit/Widgets/Battery.swift
+++ b/Kit/Widgets/Battery.swift
@@ -407,18 +407,17 @@ public class BatteryWidget: WidgetWrapper {
         }
         
         if updated {
-            self.needsDisplay = true
             DispatchQueue.main.async(execute: {
-                self.display()
+                self.renderToLayer()
             })
         }
     }
-    
+
     // MARK: - Settings
-    
+
     public override func settings() -> NSView {
         let view = SettingsContainerView()
-        
+
         var additionalOptions = BatteryAdditionals
         if self.title == "Bluetooth" {
             additionalOptions = additionalOptions.filter({ $0.key == "none" || $0.key == "percentage" })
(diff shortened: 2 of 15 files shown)


zevvgenz:
Confirming this reproduces on new hardware — **MacBook Pro M5 Max (Mac17,6), macOS 26.5.1 (25F80)**, Stats v3.0.4 (Homebrew cask).

WindowServer logs were filling with `_CGXPackagesSetWindowConstraints: Invalid window` at a steady rate. Isolated it to Stats with a clean A/B test:

| State | `Invalid window` errors / 5s |
|---|---|
| Stats running | **21** |
| `kill -STOP <stats_pid>` | **1** |
| `kill -CONT <stats_pid>` | **21** |

That's ~290k errors/day, ~95% of all WindowServer `Invalid window` noise on the machine, scaling with the menu-bar widget update interval — matches the root cause described here. Quitting Stats drops it to baseline (~1 per 10s) immediately.

Looking forward to this getting merged. 👍

stevenyu113228:
Great analysis on the `setNeedsDisplayInRect:` → replicant chain — I can confirm it's real. I built this PR locally on Tahoe 26.3.1 (3-display M1 Pro) and measured it with an lldb breakpoint on `SLSPackagesSetWindowConstraints` inside the Stats process (zero-noise measurement, immune to other apps' WindowServer errors):

| build | breakpoint hits / 60s |
|---|---|
| master | 153 |
| this PR | **46** |

So the layer-contents pipeline removes ~70% of the round-trips, but a second trigger chain remains: 45 of the 46 hits came from

```
NSISEngine layout pass → -[NSWindow(NSFullScreen) _refreshTilingConstraints:]
  → _doUpdateTilingConstraintsImmediately → SLSPackagesSetWindowConstraints
```

i.e. `setWidth()` still calls `setFrameSize`/`NSStatusItem.length`, and any autolayout pass in a status item re-schedules the tiling-constraints sync. The gate for that sync (`_needsTilingConstraintUpdate`) returns `true` whenever the app is inactive (disassembly-verified), which for a menu bar app is essentially always — so it fires regardless of how the view got invalidated.

I've proposed a complementary minimal fix that closes that gate for `NSStatusBarWindow` only (it can never participate in full-screen tiling): breakpoint hits go to 0/60s. The two approaches compose — yours removes wasted redraw work, the gate override guarantees nothing reaches WindowServer.

john-unger:
will this also fix the issue described here? https://github.com/exelban/stats/issues/3322

also there are conflicts that need to be resolved for this to be merged (I'm just waiting for this to roll out:)

PR #3219, opened 2026-05-18
Fix updater app copy preserving framework symlinks

### LINKED ISSUE #3218: Updater can corrupt framework symlinks and make Stats.app fail Gatekeeper
**Describe the bug**
The self-updater can install a copy of `Stats.app` that macOS will no longer open. After updating to v2.12.14, `open /Applications/Stats.app` failed and Gatekeeper rejected the installed app:

```text
/Applications/Stats.app: rejected (bundle format is ambiguous (could be app or framework))
```

**Root cause**
`Kit/scripts/updater.sh` currently installs the mounted app with:

```bash
cp -rf "$APP_SRC" "$APP_DST"
```

On macOS, this dereferences symlinks inside framework bundles copied from the DMG. For example, the source DMG has:

```text
Sensors.framework/Sensors -> Versions/Current/Sensors
Sensors.framework/Resources -> Versions/Current/Resources
```

After the updater-style `cp -rf` copy, those entries become a real executable file and a real directory. That breaks the framework bundle layout and makes `spctl` reject the app as ambiguous.

**Reproduction**
On macOS 26.4.1 (25E253), using the v2.12.14 DMG:

```bash
hdiutil attach -nobrowse -readonly -mountpoint /tmp/stats-mount Stats.dmg
cp -rf /tmp/stats-mount/Stats.app /tmp/stats-cp/Stats.app
spctl --assess --type execute --verbose=4 /tmp/stats-cp/Stats.app
```

Result:

```text
/tmp/stats-cp/Stats.app: rejected (bundle format is ambiguous (could be app or framework))
```

Copying the same app with `ditto` preserves the framework symlinks and passes Gatekeeper:

```bash
ditto /tmp/stats-mount/Stats.app /tmp/stats-ditto/Stats.app
spctl --assess --type execute --verbose=4 /tmp/stats-ditto/Stats.app
```

Result:

```text
/tmp/stats-ditto/Stats.app: accepted
source=Notarized Developer ID
```

**Details**
- Device: Apple Silicon Mac
- macOS: 26.4.1 (25E253)
- Application version: 2.12.14

I will link a PR that replaces the updater copy step with `ditto` so the installed app keeps the framework symlink layout from the DMG.

constansino:
Fixes #3218.

## Summary
- replace the updater's `cp -rf` app copy with `/usr/bin/ditto`
- share the app install step between both updater launch paths
- preserve framework bundle symlinks from the mounted DMG so Gatekeeper keeps accepting the installed app

## Verification
- `bash -n Kit/scripts/updater.sh`
- Ran `Kit/scripts/updater.sh` against a fake `Stats.app` containing framework-style symlinks and confirmed `Test.framework/Test`, `Test.framework/Resources`, and `Test.framework/Versions/Current` remain symlinks after install.
- Reproduced the released v2.12.14 DMG behavior locally: `cp -rf` copy is rejected by `spctl` with `bundle format is ambiguous`, while a `ditto` copy is accepted as `Notarized Developer ID`.

### DIFF
--- a/Kit/scripts/updater.sh
+++ b/Kit/scripts/updater.sh
@@ -30,9 +30,13 @@ launch_app() {
     fi
 }
 
-if [[ "$STEP" == "2" ]]; then
+install_app() {
     rm -rf "$APP_DST"
-    cp -rf "$APP_SRC" "$APP_DST"
+    /usr/bin/ditto "$APP_SRC" "$APP_DST"
+}
+
+if [[ "$STEP" == "2" ]]; then
+    install_app
 
     launch_app "$APP_DST/Contents/MacOS/Stats" --dmg "$DMG_PATH"
 
@@ -44,8 +48,7 @@ elif [[ "$STEP" == "3" ]]; then
 
     echo "Done"
 else
-    rm -rf "$APP_DST"
-    cp -rf "$APP_SRC" "$APP_DST"
+    install_app
 
     launch_app "$APP_DST/Contents/MacOS/Stats" --dmg-path "$DMG_PATH" --mount-path "$MOUNT_PATH"
 


exelban:
hi, thanks for you contribution. LGTM, but I have a few comments, maybe it's good to use `ditto` instead of `/usr/bin/ditto` just in case someone has different configuration etc. Second, because you move that call to the separate function please also add a check if `ditto` is available -> if yes, otherwise use `cp -rf`

PR #3229, opened 2026-05-21
Toggle combined details with Option key

### LINKED ISSUE #3223: [Feature Request] Right-click or hold Shift to show combined details in the Combine Modules feature
We currently have the Combine Modules feature, but there are only two options:

1. Show combined details
2. Do not combine details

I would like to be able to view both modes at the same time.

My suggestion is to add an interaction such as right-clicking or holding Shift on a combined module to temporarily show the combined details. This would allow users to keep the default combined view while still being able to inspect the detailed breakdown when needed.

IonBazan:
Allows toggling between "combined details" view using Option (⌥) key as proposed in #3223.
I decided to use Option key instead of Shift/Right click as it keeps it consistent with other native Mac OS widgets (Wi-Fi).

Note that this feature works by inverting the behaviour selected in "combined details" depending on key combination.

Fixes #3223

### DIFF
--- a/Kit/module/widget.swift
+++ b/Kit/module/widget.swift
@@ -151,6 +151,7 @@ extension widget_t: CaseIterable {}
 public protocol widget_p: NSView {
     var widthHandler: (() -> Void)? { get set }
     var onClick: (() -> Void)? { get set }
+    var onOptionClick: (() -> Void)? { get set }
     
     func settings() -> NSView
 }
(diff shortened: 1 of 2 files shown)


exelban:
about this one, the realization maybe ok but, I'm not sure this feature will be added to the Stats.
Will take a look at free time, but for sure not in the nearest feature

kk1987:
A macOS 27 note for whenever this gets rebased: inside the status-item button action, the event from NSApp.currentEvent arrives with modifierFlags stripped (type is still leftMouseDown, flags 0x0 while Option is held), so the Option check never fires. Reading the live keyboard state via the static NSEvent.modifierFlags property works on both 26 and 27 — same idea as the cursor-position module resolution master uses since aff31754.

I'm carrying this PR (authorship kept) with that adaptation in my fork, verified on 26.5 and 27.0: https://github.com/kk1987/stats/commit/764b087c

PR #3230, opened 2026-05-21
Allow selecting different calendars in clock widget

### LINKED ISSUE #2590: Feature Request: Basic math in Clock format
I would like to request the ability to add to/ subtract from a numerical value.
The reason for this is that I would like to display a non-Gregorian Calendar in my menu bar. This means that the year would not be 2025. 
There's a few countries (Japan and Taiwan) that use 'Eras' for example.
An easy and somewhat flexible way to have the Era year displayed would be to just do something like `{yyyy-1234}` (substituting yyyy for whatever number).

IonBazan:
This PR aims to fix #2590 by allowing to customize calendar to be used in clock widget. 

Available options vary between OS versions though.

image

### DIFF
--- a/Modules/Clock/main.swift
+++ b/Modules/Clock/main.swift
@@ -19,8 +19,35 @@ public struct Clock_t: Codable {
     public var name: String
     public var format: String
     public var tz: String
+    public var calendar: String
     
     public var value: Date? = nil
+    
+    private enum CodingKeys: String, CodingKey {
+        case id, enabled, name, format, tz, calendar, value
+    }
+    
+    public init(id: String = UUID().uuidString, enabled: Bool = true, name: String, format: String, tz: String, calendar: String = "gregorian", value: Date? = nil) {
+        self.id = id
+        self.enabled = enabled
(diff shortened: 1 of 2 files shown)


exelban:
please explain me how its fixes #2590

PS: i'm not telling that its a bad PR, just not related to the math in the time

IonBazan:
The initial proposal was to allow subtracting a specific number from year (for Japan it would be 2019 to mark Reiwa era). However, I believe it is better to leave date format calculation to the OS which already supports a bunch of different calendars, including these with different dates and months - some calendars don't start on 1 Jan. It won't allow counting years from any arbitrary date as it's limited to a few custom ones though. Both calendars: Japanese ([Wareki](https://en.wikipedia.org/wiki/Japanese_era_name)) and Taiwanese ([Minguo](https://en.wikipedia.org/wiki/Republic_of_China_calendar)) are supported by this change.


Mac settings allow to customize it in similar fashion: 
image

exelban:
LGTM, thanks

Issue #3233, opened 2026-05-22
Add real-time SSD I/O activity indicator (read/write throughput)


Serj92:
Thank you for your beautiful app. I have a proposition of new functionality on it.

### Summary
The current SSD indicator shows disk capacity usage. Add a separate indicator that visualizes live read/write throughput instead.

### Proposed behavior
Menu bar icon (barrel/cylinder) split horizontally at the midpoint.
Upper half represents write activity. 100% write throughput fills the entire upper half.
Lower half represents read activity. 100% read throughput fills the entire lower half.
Both halves update independently and in real time.
At rest (no I/O), the barrel is empty. At max simultaneous read and write, the barrel is fully filled.

### Rationale
Capacity fill is a static, low-information metric on modern SSDs. Real-time I/O is what users actually need to diagnose system slowdowns, background processes, Spotlight indexing, Time Machine, etc. Splitting read and write into one symmetric indicator preserves menu bar space while exposing two distinct values.

### Scaling note
Define the 100% reference for read and write (e.g. theoretical interface max, observed peak over session, or user-configurable ceiling). Make this configurable in preferences.

Thank you.

exelban:
it's already possible with current widgets

Serj92:
> it's already possible with current widgets

Is this possible in the form of vertical bars, as I described? Could you tell me how to do this? I can't find it. Thank You!

exelban:
you can have a dot/arrow/symbols for IO in the speed widget

Serj92:
@exelban I made these changes in my fork of your app. Should I share these changes with your version?

PR #3238, opened 2026-05-24
feat: split-bar I/O mode for Disk and Network widgets

Serj92:
## What
Adds an opt-in **split rendering mode** to the shared `BarChart` widget. When enabled on the Disk or Network module, the bar is divided horizontally: the lower half fills bottom-up with input throughput (read / download) and the upper half stacks output throughput (write / upload) on top.

<p align="center">Stats menu bar with split I/O on Network (download blue / upload red) and Disk (read blue / write red)</p>
<p align="center"><sub><strong>NET</strong> = split download/upload &nbsp;·&nbsp; <strong>SSD</strong> = split read/write &nbsp;·&nbsp; (RAM/CPU widgets show unrelated features from sibling PRs)</sub></p>

## Why
Today, getting both directions of I/O in the menu bar requires either two separate widgets (which doubles real-estate cost) or a `SpeedWidget` (which is text-only). For users who want a glanceable bar — one that shows *both* read & write activity in a single compact column — there's no option. Several requests touch this gap; #3233 is one.

## How it works
- Each half is scaled against half the widget height, so both at 100% exactly fill the bar.
- **Self-calibrating:** a rolling per-session peak (`peakInput`, `peakOutput`) drives the 100% reference, so the widget produces sensible visuals across spinning HDDs, NVMe drives, 100 Mbit links and 10 Gbit links without user-tuned thresholds.
- Hooked into Disk (`d.activity.read` / `d.activity.write`) and Network (`download` / `upload`) via a new `BarChart.setSplitValue(input:output:)` entry point.
- The toggle (`Split read/write` for Disk, `Split download/upload` for Network) is registered **only** in the Disk and Network widget settings — it does not appear in CPU, RAM, etc., where it has no meaning.

## Behavior unchanged unless opted in
The mode is off by default. Existing Disk/Network bar-chart users see no visual change until they enable the new toggle.

## Known trade-offs (honest list)
- **Peaks are session-local** — they reset on app restart. This was a deliberate simplicity choice (no persistence, no recalibration logic). Happy to add persistence or a user-configurable cap if you'd prefer.
- Colors are hardcoded to `systemBlue` (input) and `systemRed` (output) to match the existing `SpeedWidget` convention. Could be promoted to settings later.

## Localizations
Two new strings (`Split download/upload`, `Split read/write`) added to `en.lproj`, `ru.lproj`, `uk.lproj`. The remaining 28 language files received an English fallback via `Kit/scripts/i18n.py fix` so the i18n CI check passes — translators can fill these in later via the normal flow.

Refs #3233.

### DIFF
--- a/Kit/Widgets/BarChart.swift
+++ b/Kit/Widgets/BarChart.swift
@@ -15,13 +15,21 @@ public class BarChart: WidgetWrapper {
     private var labelState: Bool = false
     private var boxState: Bool = true
     private var frameState: Bool = false
+    private var splitState: Bool = false
     public var colorState: SColor = .systemAccent
     private var colors: [SColor] = SColor.allCases
-    
+
     private var _value: [[ColorValue]] = [[]]
     private var _pressureLevel: RAMPressure = .normal
     private var _colorZones: colorZones = (0.6, 0.8)
-    
+
+    // Rolling session peaks — used as the 100% reference in split I/O mode so
+    // the bar self-calibrates without user configuration.
+    private var peakInput: Int64 = 1
+    private var peakOutput: Int64 = 1
+
+    public var isSplitMode: Bool { self.splitState }
+
     private var boxSettingsView: NSSwitch? = nil
     private var frameSettingsView: NSSwitch? = nil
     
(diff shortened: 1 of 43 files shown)


exelban:
please, read the FAQ before.

